PR #1, opened 2015-08-26
Linked in README are treated as relative and thus broken.

technion:
Enforce absolute URL. This ensures the links in the README are clickable from Github.


unixcharles:
Thanks @technion

PR #3, opened 2015-09-02
Corrections/improvements to README.md

mjtko:
Modified example endpoint to an operational boulder demo instance.  Use the correct private key (`certificate_private_key`) for signing the CSR.


unixcharles:
Make sense :+1:

PR #8, opened 2015-10-27
Array.wrap is rails specific.

technion:
Outside of Rails, I'm seeing failures like this:

    > registration = client.register(contact: 'mailto:xxxx@lolware.net')
    NoMethodError: undefined method `wrap' for Array:Class
    from acme-client/lib/acme/client.rb:20:in `register'

As far I understand, a single contact address should be supplied, so this should be accurate.

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -17,7 +17,7 @@ class Acme::Client
 
   def register(contact:)
     payload = {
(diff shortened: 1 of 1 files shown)


unixcharles:
Make sense.

I'll will have to make it so that you can provide more than one eventually.

PR #11, opened 2015-11-08
Remove obsolete urlsafe_base64 gem

Nikita-V:


unixcharles:
Unfortunately, I don't remember exactly why I did this but I'm pretty sure it was needed  somehow.

Maybe its not worth adding the dependancy, we could easily back port the functionality.

Feel free to make a PR for this. :)

PR #17, opened 2015-12-01
Add some more steps to the example in the README

dkam:
Thanks for writing this library.   Not sure if this is the direction you want to go, but I've added some steps to save the challenge file and serve it, save the server key and certificate and start a ruby server with the key and certificate. 

I found them useful to get an actual working key and certificate working.


unixcharles:
Glad it was useful to you, thanks for contributing back.

PR #18, opened 2015-12-04
Download and return chain in Acme::Client#new_certificate

jhass:
Here's a first rough stab at providing the full certificate chain to the caller.

I didn't spend too much thought on whether this is a good API, since I'd like to discuss what the best way to provide this is with you.


unixcharles:
Its definitively a needed feature.

I term of API, I would rather not have the `#new_certificate` returning an array.

I'm thinking that it would make sense to wrap it into an dedicated object that wrap the certificate and embed the chain.

That would also be a nice entry point for providing utility method on top of `OpenSSL::X509::Certificate` in the future. Given how weird that thing is to work with, a good bit of abstraction won't hurt.

jhass:
Added a very basic wrapper so we have something to work with. Maybe it should keep track of the private key to and perhaps even the CSR? Though at least the former would make the private key a required argument to `#new_certificate`.

lib/acme/certificate.rb
```diff
@@ -0,0 +1,20 @@
...
+    @x509_chain = chain
+  end
+
+  def chain_to_pem
+    x509_chain.map(&:to_pem).join
+  end
+
+  def fullchain_to_pem
+    [*x509_chain, x509].map(&:to_pem).join
+  end
```
unixcharles: :+1: thats pretty much what people will be looking for most of the time.

README.md
```diff
@@ -80,11 +80,13 @@ csr.public_key = certificate_private_key.public_key
 csr.sign(certificate_private_key, OpenSSL::Digest::SHA256.new)
 
 # We can now request a certificate
-https_cert = client.new_certificate(csr) # => #<OpenSSL::X509::Certificate ....>
+https_cert = client.new_certificate(csr) # => #<Acme::Certificate ....>
```
unixcharles: Nitpicking but I wouldn't call it `https_cert` given that its just a certificate that would work any type of tls backed service

jhass:
Updated.

jhass:
Thanks for merging1

unixcharles:
released in `0.2.0`

PR #20, opened 2015-12-04
dns-01 challenge

jhass:
I took @Nikita-V's branch, rebased it and added some basic specs.

The cassettes were obtained using the latest boulder, slightly modified so that that some unfinished internal validation wouldn't fail, then the actual hostname was changed to `test.example.com`.

While `dns-01` is not enabled yet on LE's servers and as pointed out above the integration into boulder is unfinished when it comes down to some additional validations, I think it's worth to explore client support for this very useful challenge as early as possible, and I'd like to avoid done work towards it to rot away ;)

I'm happy to do any API changes that you like to have.


Nikita-V:
@jhass I've updated the branch dns_challenge. Can you add your specs to the new version of the branch?

jhass:
Cool, done.

unixcharles:
Extracted the CI change in #22 just to keep things organized.

spec/spec_helper.rb
```diff
@@ -48,6 +48,7 @@ def serve_once(body)
     yield
   ensure
     server.shutdown
+    thread.join(5)
     thread.exit
```
unixcharles: I assume it won't make sense to exit anymore since it would be a dead thread after calling `#join`?

unixcharles:
Awesome work :+1: 

rebased in https://github.com/unixcharles/acme-client/pull/23

spec/spec_helper.rb
jhass: No the `5` makes it return after five seconds regardless. I observed a race condition with `thread.exit` causing this to quit so fast that the next `serve_once` would find to bind to the port since it hasn't been freed yet, since the shutdown wasn't completed yet I guess. However on 2.1 `#join` would never return for some reason, this seems to work around both conditions so far.

PR #21, opened 2015-12-04
Replace domains with RF6761 6.5 domains

jhass:
Examples shouldn't refer to potentially valid domain names,
not only because of potential conflicts or no guarantees on
real world values, but also to avoid to potentially reference
undesirable content hosted on those.


unixcharles:
Make sense :+1:

PR #22, opened 2015-12-04
Fix issue with ci

unixcharles:


unixcharles:
Extracted from https://github.com/unixcharles/acme-client/pull/20/files

PR #26, opened 2015-12-04
Provide a convenience wrapper for CSR generation

### LINKED ISSUE #24: CSR generator
A very common task when using this library will be generating a CSR. Ruby's OpenSSL API is rather inconvenient, especially when it comes down to including a `subjectAltName` extension.

Should we provide a convenience class to ease CSR generation? What would the interface look like? First thoughts would be to keep it as simple as `client.new_certificate(%w(example.org www.example.org), private_key)`, `client.new_certificate(Acme.generate_csr(%w(example.org www.example.org), private_key)` or `client.new_certificate(Acme::Csr.new(%w(example.org www.example.org), private_key))`. The first hostname would be picked for the `CN`.

jhass:
closes #24

Not sure about what subject attributes to support, I kept the list rather minimal for now, most CAs reject everything but the common name anyway.

### DIFF
--- a/lib/acme/certificate_request.rb
+++ b/lib/acme/certificate_request.rb
@@ -1,0 +1,113 @@
+class Acme::CertificateRequest
+  extend Forwardable
+
+  DEFAULT_KEY_LENGTH = 2048
+  DEFAULT_DIGEST = OpenSSL::Digest::SHA256
+  SUBJECT_KEYS = {
+    common_name:         "CN",
+    country_name:        "C",
+    organization_name:   "O",
+    organizational_unit: "OU",
+    state_or_province:   "ST",
+    locality_name:       "L"
+  }.freeze
+
+  SUBJECT_TYPES = {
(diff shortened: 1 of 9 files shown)


lib/acme/certificate_request.rb
```diff
@@ -0,0 +1,113 @@
...
+  def generate_subject
+    OpenSSL::X509::Name.new(
+      @subject.map {|name, value|
+        [name, value, SUBJECT_TYPES[name]]
+      }
+    )
+  end
+
+  def add_extension(csr)
+    return if @names.size <= 1
```
unixcharles: Do you want `@names.empty?`, why return on 1?

unixcharles: oh right, then you would only have a common names?

jhass: Yes. After some tinkering I found it less confusing if `Acme::CertificateRequest.new(common_name: "example.org").names #=> ["example.org"]`

PR #27, opened 2015-12-07
Do not reimplement JWS

jhass:
json-jwt already provides it.


unixcharles:
Good stuff! Is there a reason to re-generate the vcr cassettes?

jhass:
Well, the new implementation doesn't use `header` anymore but puts all headers into `protected`. So while I was running a test against boulder anyway, I thought it might be good to get them updated too. They're actually also a nice reference for developers of other client apps :) Oh and it also solves a todo I have in #26

PR #31, opened 2015-12-07
Various refactoring

unixcharles:
ping @jhass


spec/certificate_request_spec.rb
```diff
@@ -1,22 +1,20 @@
 require 'spec_helper'
 
 describe Acme::CertificateRequest do
-  before(:all) do
-    @test_key = generate_private_key
-  end
+  let(:test_key) { generate_private_key }
```
jhass: This seem to indeed make the specs on average 3 (2-4) seconds slower on my machine.

lib/acme/crypto.rb
```diff
@@ -13,14 +13,16 @@ def generate_signed_jws(header:, payload:)
...
 
-  def jwk
-    @jwk ||= JSON::JWK.new(public_key)
-  end
-
   def thumbprint
     jwk.thumbprint
   end
 
+  private
```
jhass: I guess we miss a spec calling this https://github.com/unixcharles/acme-client/blob/master/lib/acme/resources/challenges/dns01.rb#L14

spec/certificate_request_spec.rb
```diff
@@ -1,22 +1,22 @@
 require 'spec_helper'
 
 describe Acme::CertificateRequest do
-  before(:all) do
-    @test_key = generate_private_key
+  def test_key
+    @test_key ||= generate_private_key
   end
```
jhass: RSpec seems to do some magic for `before(:all)`, inspecting it, each example still resets the instance variable this way.

unixcharles: juste moved it back into before(:all) and trimmed few more second by only generating the CSR when calling it.

jhass:
LGTM so far :+1:

spec/spec_helper.rb
```diff
@@ -1,63 +1,19 @@
 $LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
+$LOAD_PATH.unshift File.expand_path('../support', __FILE__)
```
jhass: Oh, just realized, since we only support 2.1 and 2.2, we could use `__dir__` here and trim one `../` from each. YMMV.

unixcharles: :+1:

PR #32, opened 2015-12-07
introduce keystash, reuse cached keys in specs

jhass:
Okay this is a bit controversial, but it brings running the entire testsuite below one second for me.

This introduces a cache for generated private keys. If you need fresh keys for example to generate a fresh cassette, simply delete `spec/fixtures/keystash.yml`. If you add a new spec that needs more keys, the keystash will automatically append new ones. However this does not guarantee the same examples will get the same keys, they will randomize among the cache. This works in stubbed mode, since there the responses are all fake anyhow. There also doesn't seem to be an issue for VCR to match up the requests (well, this didn't change anything about it, as previously new keys were generated all the time).

The perhaps biggest concern is bloating the repo with another rather big (and over time only growing) fixture.

Excuse the drive by changes back filling and improving a couple of specs.


spec/authorization_spec.rb
```diff
@@ -11,8 +11,14 @@
...
 
   context '#http01' do
-    it 'returns a HTTP01 object', vcr: { cassette_name: 'authorization_http_01' }  do
+    it 'returns a HTTP01 object', vcr: { cassette_name: 'authorization' }  do
       expect(authorization.http01).to be_a(Acme::Resources::Challenges::HTTP01)
     end
   end
+
+  context '#dns01' do
+    it 'returns a DNS01 object', vcr: { cassette_name: 'authorization' }  do
```
unixcharles: Are you sure you want to share the cassette for both scenario here?

spec/support/ssl_helper.rb
```diff
@@ -1,9 +1,41 @@
...
+      @keystash = load
+      @iter = @keystash.each
+    end
+
+    def next
+      @iter.next
+    rescue StopIteration
+      @keystash << OpenSSL::PKey::RSA.new(2048)
+      save
+      @keystash.last
```
unixcharles: I like that it can generate itself

spec/authorization_spec.rb
jhass: Yes, both cassettes would contain the same stuff since the server responds with all challenges available.

unixcharles: :+1:

jhass:
Alright, adventure accepted I see :D

unixcharles:
its kind of funky but i like it :)

jhass:
Yeah, it should be easy enough to rip out again if it creates too much of a mess anyhow.

unixcharles:
In a way its pretty similar to cassettes, you just delete it when you want new ones

PR #33, opened 2015-12-10
Prepare dropping activesupport

jhass:
I'm trying to get rid of activesupport upstream, see https://github.com/nov/json-jwt/pull/29

This should be merged as that's released.

Both `Object.const_get` and `Object.const_defined?` resolve `::`s in their argument these days.

### DIFF
--- a/lib/acme/client/faraday_middleware.rb
+++ b/lib/acme/client/faraday_middleware.rb
@@ -36,9 +36,9 @@ class Acme::Client::FaradayMiddleware < Faraday::Middleware
 
     return if env.success?
 
-    error_name = env.body['type'].gsub('urn:acme:error:', '').classify
-    error_class = if Acme::Client::Error.qualified_const_defined?(error_name)
-      "Acme::Client::Error::#{error_name}".constantize
+    error_name = env.body['type'].gsub('urn:acme:error:', '').split(/[_-]/).map(&:capitalize).join
+    error_class = if Acme::Client::Error.const_defined?(error_name)
+      Object.const_get("Acme::Client::Error::#{error_name}")
     else
       Acme::Client::Error
     end


unixcharles:
:+1: thanks

its kind of nasty that I was using activesupport without explicitly declaring it

elijh:
yes, please!

jhass:
Upstream ignores so if we want this we have to move away from json-jwt.

will-in-wi:
Is https://github.com/jwt/ruby-jwt a substitute?

jhass:
Possibly, yes. I don't know if they have feature parity.

kyledrake:
Any updates on this? Activesupport is doing some really "fun" things to my site.

cancer

jhass:
If you manage to port the gem to `ruby-jwt` instead, that has good chances to be accepted, I'm all in favour at least. @nov completely ignores us here.

unixcharles:
Yes, I would be open to use ruby-jwt instead 👍

kyledrake:
ruby-jwt apparently doesn't implement jwk. They mentioned possibly implementing it in a 2.0 branch.. that doesn't exist.

I'm probably just going to fork it and jerry rig a couple custom branches for now to move forward, I don't really know what else to do here. I'll try emailing @nov and see if I can get ahold of him.

shaneshort:
has anyone considered forking json-jwt and implementing this fix?

unixcharles:
That would be an option on the short term

jhass:
The main issue with forking a dependent gem as a gem is that you need to publish your fork to rubygems.org, possibly creating confusion in the community about what it is.

kyledrake:
Thank you for this! I'm looking forward to a release, gets rid of a lot of hacky code I used to get around activesupport issues.

unixcharles:
thanks to @jhass

PR #34, opened 2015-12-12
duplicated range & unused variable

manuel-io:
acme-client/lib/acme/faraday_middleware.rb:30:
  warning: character class has duplicated range: /rel="([\w_-]+)"/

\w : Any word character (letter, number, underscore)

### DIFF
--- a/lib/acme/faraday_middleware.rb
+++ b/lib/acme/faraday_middleware.rb
@@ -27,7 +27,7 @@ class Acme::FaradayMiddleware < Faraday::Middleware
       link_header = env.response_headers['Link']
       links = link_header.split(', ').map do |entry|
         link = entry.match(/<(.*?)>;/).captures.first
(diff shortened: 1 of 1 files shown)


lib/acme/faraday_middleware.rb
```diff
@@ -43,7 +43,7 @@ def on_complete(env)
       Acme::Error
     end
 
-    message = if env.body.is_a? Hash
```
jhass: Maybe this meant to be used in the `raise` below?

unixcharles: Yes it was. I'll fix that.

unixcharles:
fixed in https://github.com/unixcharles/acme-client/pull/35

lib/acme/faraday_middleware.rb
manuel-io: My fail :-(

PR #36, opened 2015-12-13
Support for tls-dvsni-01

saltsa:
sni_name method returns the server name to be used in webserver config and to create self-signed certificate


jhass:
I'm not sure we should use the term `dvsni` anywhere, there's no single mention of it in https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md

I realize that's a bit more verbose, but how about `TLSSNI01` for the class and `tls_sni_01` for the method name?

lib/acme/resources/challenges/dvsni01.rb
```diff
@@ -0,0 +1,12 @@
+class Acme::Resources::Challenges::DVSNI01 < Acme::Resources::Challenges::Base
+
+  def sni_name
+    hd = crypto.digest.hexdigest(authorization_key)
+    "#{hd[0..31]}.#{hd[32..64]}.acme.invalid"
+  end
+
+  def request_verification
+    response = client.connection.post(@uri, { resource: 'challenge', type: 'dvsni-01', keyAuthorization: authorization_key })
```
jhass: Did you test this? The old and removed name was `dvsni`: https://github.com/letsencrypt/boulder/pull/1247/files#diff-8be8b6353017e06063c146243e3eb2d2L71 but I can't find `dvsni-01` anywhere. The new name is `tls-sni-01`: https://github.com/letsencrypt/boulder/blob/6a575430f1cd4a27a498937a4fd12d258fe26a7c/core/objects.go#L71 and https://github.com/letsencrypt/boulder/blob/6a575430f1cd4a27a498937a4fd12d258fe26a7c/va/validation-authority.go#L585-L586

saltsa: Sorry, messed up with the commits. Yes, indeed it's tls-sni-01.

jhass:
I think generating the actual certificate will be a very common task, maybe we can provide some convenience API for it? I'm thinking along the lines of

```
challenge.certificate # => Acme::Certificate 
challenge.certificate.private_key #=> OpenSSL::RSA::PKey
```

The convenience method for generating the name should stay of course and certificate generation would be lazy until first access, in case one wants to do it manually instead. `Acme::Certificate`s new `private_key` would be something like `@private_key || (@request && @request.respond_to?(:private_key) && @request.private_key)`.

saltsa:
I could try to integrate self-sign generating code which I have implemented outside this repo.

lib/acme/resources/challenges/dvsni01.rb
```diff
@@ -0,0 +1,12 @@
+class Acme::Resources::Challenges::DVSNI01 < Acme::Resources::Challenges::Base
+
+  def sni_name
```
jhass: That a certificate is returned on the TLS server when asked for the SNI name here is only half of the challenge, that the certificate also includes that name as a SAN is equally important. I wonder if we can find a name for this method that conveys both facts or at least doesn't mention one while it doesn't mention the other.

`challenge_hostname` or just `hostname` perhaps?

lib/acme/resources/challenges/base.rb
```diff
@@ -1,6 +1,6 @@
 class Acme::Resources::Challenges::Base
 
-  attr_reader :client, :status, :uri, :token, :error
+  attr_reader :client, :status, :uri, :token, :error, :iterations
```
jhass: Leftover?

jhass:
> I could try to integrate self-sign generating code which I have implemented outside this repo.

Sounds good to me, @unixcharles WDYT?

lib/acme/resources/challenges/base.rb
saltsa: Yes, sorry, have to learn more about these Github PR. That was fixed in c285955

lib/acme/resources/challenges/dvsni01.rb
```diff
@@ -0,0 +1,12 @@
...
+    hd = crypto.digest.hexdigest(authorization_key)
+    "#{hd[0..31]}.#{hd[32..64]}.acme.invalid"
+  end
+
+  def request_verification
+    response = client.connection.post(@uri, { resource: 'challenge', type: 'tls-dvsni-01', keyAuthorization: authorization_key })
```
jhass: It's `tls-sni-01`, not `tls-dvsni-01`. Honestly I think you need a test environment :/

saltsa: Well, I'm testing against the acme-staging server and that seems to work. It seems it doesn't check the "type" field at all and does the validation still against the self-signedSNI certificate?!

jhass: That's hard to believe.

lib/acme/resources/challenges/base.rb
jhass: Well, something reintroduced it, https://github.com/saltsa/acme-client/commit/b465740a72758c88d23d12f534eb322d75ec79ee perhaps?

lib/acme/resources/challenges/tlssni01.rb
```diff
@@ -0,0 +1,12 @@
+class Acme::Resources::Challenges::TLSSNI01 < Acme::Resources::Challenges::Base
+
+  def sni_name
```
jhass: That a certificate is returned on the TLS server when asked for the SNI name here is only half of the challenge, that the certificate also includes that name as a SAN is equally important. I wonder if we can find a name for this method that conveys both facts or at least doesn't mention one while it doesn't mention the other.

`challenge_hostname` or just `hostname` perhaps?

jhass:
Since we can do the convenience API for generating the certificate also in a follow up PR, the only real merge blocker here for me are the missing tests.

saltsa:
I integrated the self-signed cert and key generation code. Those can be fetched directly as PEM-format, like this:

```
    sni_chal = authorization.tlssni01
    puts "SNI domain: #{sni_chal.hostname}"
    tmp_key = sni_chal.self_signed_certificate_key
    tmp_cert = sni_chal.self_signed_certificate
```

I did test this against staging-server and seems to work. Sorry about the lack of any tests...

lib/acme/client.rb
```diff
@@ -47,6 +49,12 @@ def new_certificate(csr)
     ::Acme::Certificate.new(OpenSSL::X509::Certificate.new(response.body), fetch_chain(response), csr)
   end
 
+  def new_self_signed_certificate domain
+    cert = ::Acme::Certificate.new(nil, nil, nil)
+    cert.create_self_signed_cert([domain])
```
jhass: `tls-sni-01` has the constraint of the name having to be in a cert in a SAN. This is rather untypical for a certificate only covering a single domain, and thus `Acme::CertificateRequest` for example generates only the `CN` but no SAN in the single domain case. Thus I think we should keep the logic to generate a self signed `OpenSSL::X509::Certificate` inside the challenge class and pass it to `Acme::Certificate`.

lib/acme/resources/challenges/tlssni01.rb
```diff
@@ -0,0 +1,24 @@
...
+    "#{hd[0..31]}.#{hd[32..64]}.acme.invalid"
+  end
+
+  def temporary_cert
+    @temporary_cert ||= client.new_self_signed_certificate self.hostname 
+  end
+
+  def self_signed_certificate
+    self.temporary_cert.to_pem
+  end
```
jhass: Is it worth repeating `self_signed` here or isn't it obvious from the context?
```ruby
authorization = client.authorize(domain: "example.org")
challenge = authorization.tlssni01
serve(challenge.certificate)
challenge.request_verification
certificate = client.new_certificate(Acme::CertiifcateRequest(names: ["example.org"])
```

lib/acme/certificate.rb
```diff
@@ -9,6 +9,46 @@ def initialize(certificate, chain, request)
     @x509 = certificate
     @x509_chain = chain
     @request = request
+
+    #create_self_signed_cert if x509.nil?
+  end
+
+  def create_self_signed_cert subjectAltDomains
+    keypair = OpenSSL::PKey::RSA.new(2048)
```
jhass: The private key should probably be injectable.

lib/acme/resources/challenges/tlssni01.rb
saltsa: Probably not. However, the `create_self_signed_cert` itself supports multiple hostnames: The first one is used in `CN=` and all hostnames are used as SAN names.

lib/acme/certificate.rb
```diff
@@ -9,6 +9,39 @@ def initialize(certificate, chain, request)
...
     @request = request
+
+    #create_self_signed_cert if x509.nil?
+  end
+
+  def create_self_signed_cert subjectAltDomains
+    keypair = OpenSSL::PKey::RSA.new(2048) 
+
+    cert = OpenSSL::X509::Certificate.new
+    cert.subject = cert.issuer = OpenSSL::X509::Name.parse "/CN=#{subjectAltDomains.first}"
```
jhass: This seems like a workaround in order to avoid using the proper API for this, `OpenSSL::X509::Name.new([["CN", hostname, OpenSSL::ASN1::UTF8STRING]])`

lib/acme/certificate.rb
```diff
@@ -9,6 +9,39 @@ def initialize(certificate, chain, request)
...
+    cert.not_before = Time.now
+    cert.not_after = Time.now + 48*3600
```
jhass: Two days is impractical for a general purpose API as it is now, but probably actually a bit long for a special purpose API when moved into the challenge class. It would be good to find out for how long boulder keeps a pending challenge alive.

lib/acme/certificate.rb
```diff
@@ -9,6 +9,39 @@ def initialize(certificate, chain, request)
...
+
+    #create_self_signed_cert if x509.nil?
+  end
+
+  def create_self_signed_cert subjectAltDomains
+    keypair = OpenSSL::PKey::RSA.new(2048) 
+
+    cert = OpenSSL::X509::Certificate.new
+    cert.subject = cert.issuer = OpenSSL::X509::Name.parse "/CN=#{subjectAltDomains.first}"
+    cert.not_before = Time.now
```
jhass: Is setting this explicitly really necessary? Besides that it seems to be good practice to date back a certificate a little bit (an hour or so) to accommodate for clock drift, DST changes etc.

lib/acme/certificate.rb
```diff
@@ -9,6 +9,39 @@ def initialize(certificate, chain, request)
...
+    cert.not_after = Time.now + 48*3600
+    cert.public_key = keypair.public_key
+    cert.version = 2 # X509v3
+    cert.serial = 0x0
```
jhass: Do we really need to set the version and serial explicitly?

lib/acme/certificate.rb
saltsa: Well, IMHO no. The self-signed cert is anyway temporary and you have to copy the certificate PEM to the correct location. I think, at the same cost, one can copy the private key PEM too.

lib/acme/certificate.rb
```diff
@@ -9,6 +9,39 @@ def initialize(certificate, chain, request)
...
+    ef.subject_certificate = cert
+    ef.issuer_certificate = cert
+
+    cert.extensions = [
+        ef.create_extension("basicConstraints","CA:TRUE", true),
+        ef.create_extension("subjectKeyIdentifier", "hash"),
+    ]
+
+    cert.add_extension ef.create_extension("authorityKeyIdentifier", "keyid:always,issuer:always")
+    cert.add_extension ef.create_extension("subjectAltName", subjectAltDomains.map { |d| "DNS: #{d}" }.join(','))
```
jhass: This seems a bit inconsistent, either populate `extensions=` with the full array or use `add_extension` for all four.

lib/acme/resources/challenges/tlssni01.rb
jhass: See above regarding my thoughts on the general applicability of `create_self_signed_cert`.

lib/acme/certificate.rb
jhass: I disagree, private keys usually have a higher protection than the corresponding certificate. It does seem contrived, but being able to read but not write an existing private key that must be used is not unimaginable for me. Besides the contrived real world scenario, being able to provide an existing private key will speed up the unit tests for this method by an order of magnitude easily.

lib/acme/client.rb
```diff
@@ -33,6 +33,8 @@ def authorize(domain:)
       }
     }
 
+    @domain = domain
```
jhass: I don't see this used anywhere, I guess another leftover?

lib/acme/certificate.rb
```diff
@@ -9,6 +9,39 @@ def initialize(certificate, chain, request)
...
+    cert.serial = 0x0
+
+    ext = []
+
+    ef = OpenSSL::X509::ExtensionFactory.new
+    ef.subject_certificate = cert
+    ef.issuer_certificate = cert
+
+    cert.extensions = [
+        ef.create_extension("basicConstraints","CA:TRUE", true),
```
jhass: The official client also sets `pathlen:0` here. https://github.com/letsencrypt/letsencrypt/blob/master/acme/acme/crypto_util.py#L212

lib/acme/certificate.rb
jhass: https://github.com/letsencrypt/letsencrypt/blob/master/acme/acme/challenges.py#L372 claims it's too big to fit into the CN?

lib/acme/certificate.rb
unixcharles: I'm not against autogenerating but it does make sense to allow the user to provide one.

lib/acme/resources/challenges/tlssni01.rb
```diff
@@ -0,0 +1,24 @@
+class Acme::Resources::Challenges::TLSSNI01 < Acme::Resources::Challenges::Base
+
+  def hostname
+    hd = crypto.digest.hexdigest(authorization_key)
+    "#{hd[0..31]}.#{hd[32..64]}.acme.invalid"
+  end
+
+  def temporary_cert
+    @temporary_cert ||= client.new_self_signed_certificate self.hostname
```
unixcharles: Any reason to provide a `new_self_signed_certificate` to the client instead of calling `Acme::Certificate` directly?

lib/acme/resources/authorization.rb
```diff
@@ -1,8 +1,9 @@
 class Acme::Resources::Authorization
   HTTP01 = Acme::Resources::Challenges::HTTP01
   DNS01 = Acme::Resources::Challenges::DNS01
+  TLSSNI01 = Acme::Resources::Challenges::TLSSNI01
 
-  attr_reader :domain, :status, :http01, :dns01
+  attr_reader :domain, :status, :http01, :dns01, :tlssni01
```
unixcharles: by convention i think it should be `tls_sni01`

unixcharles:
thats a pretty good start! thanks @saltsa

lib/acme/resources/challenges/tlssni01.rb
jhass: As said I think the generation code should go into the challenge code for now, see above for some reasoning about that.

lib/acme/resources/challenges/tlssni01.rb
```diff
@@ -0,0 +1,24 @@
+class Acme::Resources::Challenges::TLSSNI01 < Acme::Resources::Challenges::Base
+
+  def hostname
+    hd = crypto.digest.hexdigest(authorization_key)
+    "#{hd[0..31]}.#{hd[32..64]}.acme.invalid"
+  end
+
+  def temporary_cert
+    @temporary_cert ||= ::Acme::Certificate.new(nil, nil, nil).create_self_signed_cert([domain])
```
jhass: Really, this API is a bit terrible this way and atm nobody will find any use for using `create_self_signed_cert` in another context. Just move it into method of this class that returns an `Acme::Client::Certificate` please.

unixcharles:
Tied up a few things and added tests in https://github.com/unixcharles/acme-client/pull/41

Issue #37, opened 2015-12-13
Namespace issues


jhass:
So there's one bigger issue we maybe should discuss while we're still as early as we are currently.

The gem's name is `acme-client`, per http://guides.rubygems.org/name-your-gem/ and community conventions this gives us the namespace `Acme::Client`.

That means we do not own the namespace `Acme`. In fact other gems are already using it too, although for nothing serious as far as I can tell. However I can easily imagine other serious gems popping up in this namespace. We shouldn't use any constant below `Acme` except for `Acme::Client`.

There are four ways forward:

1. Claim this is a non-issue and that our constants inside `Acme` besides `Client` are acceptable use since generally useful to projects implementing the ACME protocol.
2. Move everything below `Acme::Client`. Currently I would only consider the rename of `Acme::CertificateRequest`  to `Acme::Client::CertificateRequest` a breaking change of public API.
3. Rename the gem into something outside of the `Acme` namespace, for example `acme_client` (`AcmeClient`, `AcmeClient::CertificateRequest`). This would involve deprecating the currently published gem and publishing a new one.
4. Ask @smathy for control over the `Acme` gem, effectively deprecating the current `acme` and replacing it with a new one. Then we could decide how to reorganize this gem, for example moving parts of it into a `acme-utils` gem or the `acme` gem itself. (I think this is the least favourable alternative at the moment, so sorry for the ping in advance @smathy).

Thoughts?

smathy:
Even if you controlled the `acme` gem that wouldn't prevent other people from creating gems (and hence constants) in the `Acme` namespace. Conversely, that I control the `acme` gem doesn't prevent you from creating any gems in the `Acme` namespace.

jhass:
Yes, but we could argue using constants in it besides `Client` better. Anyway, as I wrote I only listed it for completeness.

smathy:
:+1:

unixcharles:
sure, we could move the other classes just to be on the safe side.

PR #38, opened 2015-12-15
Move everything under the Acme::Client namespace

jhass:
Implements the first solution of #37


spec/spec_helper.rb
```diff
@@ -1,7 +1,7 @@
 $LOAD_PATH.unshift File.join(__dir__, '../lib')
 $LOAD_PATH.unshift File.join(__dir__, 'support')
 
-require 'acme-client'
+require 'acme/client'
```
unixcharles: acme-client.rb is still there?

jhass: Yes, for compatibility reasons with for example bundler's auto require. But the recommended require line is like this, see http://guides.rubygems.org/name-your-gem/

PR #39, opened 2015-12-20
add #revoke_certificate to client

sawanoboly:
Hi, I've add revoke_cert request. I'd like to revoke a certificate.

And, I think that need to ask you few questions.

- What object should this method returns?
    - F.Y.I boulder returns 200 status and empty body on success.
- Should I add spec about `#revoke_certificate` ?

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -49,6 +49,16 @@ class Acme::Client
     ::Acme::Client::Certificate.new(OpenSSL::X509::Certificate.new(response.body), fetch_chain(response), csr)
   end
 
+  def revoke_certificate(certificate)
+    payload = {
+      resource: 'revoke-cert',
+      certificate: Base64.urlsafe_encode64(certificate.to_der)
+    }
+
+    response = connection.post(@operation_endpoints.fetch('revoke-cert'), payload)
+    response.status
+  end
+
   def connection
     @connection ||= Faraday.new(@endpoint) do |configuration|
       configuration.use Acme::Client::FaradayMiddleware, client: self
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -49,14 +49,20 @@ class Acme::Client
     ::Acme::Client::Certificate.new(OpenSSL::X509::Certificate.new(response.body), fetch_chain(response), csr)
   end
 
-  def revoke_certificate(certificate)
+  def revoke_certificate(certificate, revocation_key: @private_key)
     payload = {
       resource: 'revoke-cert',
       certificate: Base64.urlsafe_encode64(certificate.to_der)
     }
 
-    response = connection.post(@operation_endpoints.fetch('revoke-cert'), payload)
-    response.status
+    if @private_key.respond_to?(:to_der) && revocation_key.to_der == @private_key.to_der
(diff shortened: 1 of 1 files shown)


jhass:
> What object should this method returns?

I think nil on success and raising on errors is okay. 

> Should I add spec about #revoke_certificate ?

Yes, please!

jhass:
One tricky problem here is that ACME allows revoking with both, the account key and the certificates private key, so we need to find a way to pass a private key down to our middleware that creates the JWS signatures alongside a request.

sawanoboly:
> I think nil on success and raising on errors is okay.

Alright, I'll update to return nil.

> ACME allows revoking with both, the account key and the certificates private key

I see. But, It is difficult to decide which class should have revoke method by private_key. Where do you think better?
Maybe, create a new resource might be good.

jhass:
As for the user facing API, I'd just do `def revoke_certificate(certificate, revocation_key: private_key)`.

sawanoboly:
I want to clarify,

When `#revoke_certificate` called without private_key, we can revoke by `self(client)` simply.
When `#revoke_certificate` called with private_key, we must create new client and call API by it for revocation.

Is it correct ?

jhass:
That would be one way. I'm not yet certain whether I would consider it cleaner than say passing a special header that's used (and removed) by the faraday middleware.

sawanoboly:
hmm... I wrote to try below. How it is ?

```
  def revoke_certificate(certificate, revocation_key: private_key)
    payload = {
      resource: 'revoke-cert',
      certificate: Base64.urlsafe_encode64(certificate.to_der)
    }

    if revocation_key.to_der == private_key.to_der
      response = connection.post(@operation_endpoints.fetch('revoke-cert'), payload)
    else
      tmp_client = ::Acme::Client.new(private_key: revocation_key, endpoint: @endpoint, directory_uri: @directory_uri)
      response = tmp_client.connection.post(@operation_endpoints.fetch('revoke-cert'), payload)
    end

    nil
  end
```

lib/acme/client.rb
```diff
@@ -49,6 +49,22 @@ def new_certificate(csr)
...
+    }
+
+    if @private_key.respond_to?(:to_der) && revocation_key.to_der == @private_key.to_der
+      response = connection.post(@operation_endpoints.fetch('revoke-cert'), payload)
+    else
+      tmp_client = ::Acme::Client.new(private_key: revocation_key, endpoint: @endpoint, directory_uri: @directory_uri)
+      response = tmp_client.connection.post(@operation_endpoints.fetch('revoke-cert'), payload)
+    end
+
+    nil
```
unixcharles: I would rather return true if it succeed, maybe return `response.success?` from faraday's response object.

This way externally you can use it in conditional.

```ruby
if client.revoke_certificate(..)
  handle_success!
else
  handle_failure!
end
```

unixcharles:
> One tricky problem here is that ACME allows revoking with both, the account key and the certificates private key, so we need to find a way to pass a private key down to our middleware that creates the JWS signatures alongside a request.

What about having `Acme::Client#revoke_certificate` on the instance method that simply revoke using the account key and then have another class method `Acme::Client.revoke_certificate` that operate on the certificate key?

sawanoboly:
close.

reason: included into #40

PR #40, opened 2015-12-20
Certificate revoke

unixcharles:
I wrote some test on top of https://github.com/unixcharles/acme-client/pull/39 and changed the API the API slightly.

Imo, I don't see too much the need for passing certificate to `revoke_certificate`, given that in that context the instance always has an account private key.

I added a class method that take client options, so its easy to do so if needed.

```ruby
client.revoke_certificate(certificate)
# => true
Acme::Client.revoke_certificate(certificate, private_key: certificate_private_key)
# => true
```

cc @sawanoboly @jhass


jhass:
Yeah I guess that works. My first intuition on preferred client API was allowing something like `client.revoke_certificate(certificate, @options[:private_key] || account_key)` but
```rb
if @options[:private_key]
  Acme::Client.revoke_certificate(certificate, @options[:private_key])
else
  client.revoke_certificate(certificate)
end
```
makes no real difference.

sawanoboly:
@unixcharles Thanks! It looks like useful than implementing into single instance method.

unixcharles:
Thanks guys for reviewing

PR #41, opened 2015-12-21
tls-sni-01

unixcharles:
This PR finish @saltsa initial implementation

- Refactoring on tls-sni-01
- simplify self-sign certificate
- Move self-sign certificate related code into its own class
- Add tests

ping @jhass @saltsa


lib/acme/client/self_sign_certificate.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def initialize(private_key: generate_private_key, subject_alt_names:)
+    @private_key = private_key
+    @subject_alt_names = subject_alt_names
+  end
+
+  def certificate
+    @certificate ||= begin
+      certificate = OpenSSL::X509::Certificate.new
+      certificate.not_before = Time.now
```
jhass: Let's back date it for an hour to account for clock drift, that seems to be common practice.

jhass:
LGTM otherwise :+1:

lib/acme/client/self_sign_certificate.rb
unixcharles: Sound reasonable

lib/acme/client/self_sign_certificate.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  private
+
+  def generate_private_key
+    OpenSSL::PKey::RSA.new(2048)
+  end
+
+  def default_not_before
+    Time.now + 3600
```
jhass: Uh, that's the wrong direction, isn't it? We want the certificate to be valid for one hour in the past.

lib/acme/client/self_sign_certificate.rb
```diff
@@ -0,0 +1,50 @@
...
+      certificate = OpenSSL::X509::Certificate.new
+      certificate.not_before = not_before
+      certificate.not_after = not_after
+      certificate.public_key = private_key.public_key
+
+      extension_factory = OpenSSL::X509::ExtensionFactory.new
+      extension_factory.subject_certificate = certificate
+      extension_factory.issuer_certificate = certificate
+
+      subject_alt_name_entry = subject_alt_names.map { |d| "DNS: #{d}" }.join(',')
```
saltsa: dunno, but maybe should remove the space after "DNS:".

lib/acme/client/self_sign_certificate.rb
unixcharles: oh, good catch

lib/acme/client/self_sign_certificate.rb
unixcharles: I think its suppose to be like that but I'll have a look

PR #43, opened 2015-12-24
Adds ruby version requirement to the readme

blackjid:
Using required keyword arguments only works from ruby 2.1 and above...

for example in here... https://github.com/unixcharles/acme-client/blob/master/lib/acme/client.rb#L12


unixcharles:
:+1: thanks

PR #45, opened 2015-12-28
Add Rubocop code style check

unixcharles:
It should not change the behaviour, just make the code base more consistent.

ping @jhass

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,79 @@
+AllCops:
+  RunRailsCops: false
+  Exclude:
+    - 'bin/*'
+    - 'vendor/**/*'
+
+Style/FileName:
+  Exclude:
+    - 'lib/acme-client.rb'
+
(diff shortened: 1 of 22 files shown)


jhass:
+1 on enforcing a consistent style, maybe enable HoundCI too. Like many people I rather dislike the default config, the [config](https://github.com/diaspora/diaspora/blob/develop/.rubocop.yml) we're using over at diaspora has some reasoning for some of the changes. But that's largely just my opinion, your project, your choices ;)

unixcharles:
> +1 on enforcing a consistent style, maybe enable HoundCI too

I find the comment slightly noisy but it will fail on travis.

> Like many people I rather dislike the default config, the config we're using over at diaspora has some reasoning for some of the changes. But that's largely just my opinion, your project, your choices ;)

I'm using the config we enforce in Shopify core + personal taste tweak but I'm open to change, let me know if something there bother you.

PR #46, opened 2016-01-05
Update dns-01 record content to comply with ACME spec

jhass:
Following implementation reality in boulder, see https://github.com/letsencrypt/boulder/pull/1295

### DIFF
--- a/lib/acme/client/resources/challenges/dns01.rb
+++ b/lib/acme/client/resources/challenges/dns01.rb
@@ -12,6 +12,6 @@ class Acme::Client::Resources::Challenges::DNS01 < Acme::Client::Resources::Chal
   end
 
   def record_content
-    crypto.digest.hexdigest(authorization_key)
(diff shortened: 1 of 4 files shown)


lib/acme/client/resources/challenges/dns01.rb
```diff
@@ -12,6 +12,6 @@ def record_type
   end
 
   def record_content
-    crypto.digest.hexdigest(authorization_key)
+    Base64.urlsafe_encode64(crypto.digest.digest(authorization_key)).sub(/[\s=]*\z/, '')
```
unixcharles: what is the sub for exactly?

jhass: It drops the padding `urlsafe_encode64` adds, which Ruby's stdlib sadly has no option for.

unixcharles: ok

dpiddy:
Would it be possible to get a gem cut with this fix included? Thanks!

unixcharles:
@dpiddy bumped to 0.3.0

PR #47, opened 2016-01-11
Reference the correct private key file

weppos:
This is a trivial change. In the usage section you write the private key in a file called `privkey.pem`

```ruby
File.write("privkey.pem", certificate.request.private_key.to_pem)
```

but then you start the server reading from `key.pem`. It's easy to figure out what's happening if you read the error message you get from the server, however it's also an easy fix. :)


unixcharles:
thanks!

PR #49, opened 2016-01-20
Expose the authorization expiration timestamp

ajacques:
The ACME server returns an optional timestamp for authorizations that signifies the expiration date of the domain authorization challenge. The time format is RFC3339 and can be parsed by Time#parse.

See: https://letsencrypt.github.io/acme-spec/
Section 5.3 - expires

Testing Done:
Updated and verified unit tests
Tested on live data with a test instance of the https://github.com/letsencrypt/boulder server

I'm currently experimenting with acme-client, and if all goes well I may have more improvements for it, but now for this is what I have.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,8 +1,11 @@
 AllCops:
-  RunRailsCops: false
+  TargetRubyVersion: 2.2
   Exclude:
     - 'bin/*'
     - 'vendor/**/*'
+
+Rails:
+  Enabled: false
 
 Style/FileName:
   Exclude:
@@ -29,7 +32,10 @@ Style/AlignParameters:
 Style/FirstParameterIndentation:
   EnforcedStyle: consistent
 
-Style/TrailingComma:
+Style/TrailingCommaInArguments:
+  Enabled: false
+
+Style/TrailingCommaInLiteral:
   Enabled: false
 
 Style/StringLiterals:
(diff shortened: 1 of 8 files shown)


jhass:
-1 on using `Time.parse`, when you do know the format, it's slow and error prone as it still does guesswork. I would suggest to check if there's a convenience method availble and fallback to `strptime` if not.

unixcharles:
Isn't RFC3339 just the same as iso8601? Time.iso8601(date) would probably just work

ajacques:
Yeah you're right on the time format, good catch not sure what I was thinking. I'll fix that

Also, the build is failing due to a Rubocop rule error. As of a recent reason the Style/TrailingComma rule has been removed as it's been replaced in exchange for two other cops. Not sure if you want it fixed in this PR or separately.

unixcharles:
> Also, the build is failing due to a Rubocop rule error. As of a recent release the Style/TrailingComma rule has been removed as it's been replaced in exchange for two other cops. Not sure if you want it fixed in this PR or separately.

Don't bother with another PR, I should have lock the Rubocop version to avoid this.

If you can just quick fix it at the same time that would be awesome :+1:

ajacques:
Well this PR kind of expanded in scope. Anyway, I fixed all of the Rubocop violations that came about from the 0.36.0 update. Plus pinned the version to 0.36.0.

unixcharles:
:+1:

PR #52, opened 2016-01-31
Fix some typos and improve copy

louim:
Hello!

I fixed some typos in the readme, and improved some comments. Cheers! :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,18 +3,21 @@
 
 `acme-client` is a client implementation of the [ACME](https://letsencrypt.github.io/acme-spec) protocol in Ruby.
 
-You can find the server reference implementation for ACME server [here](https://github.com/letsencrypt/boulder) and also the a reference [client](https://github.com/letsencrypt/letsencrypt) in python.
+You can find the ACME reference implementations of the [server](https://github.com/letsencrypt/boulder) in Go and the [client](https://github.com/letsencrypt/letsencrypt) in Python.
 
-ACME is part of the [Letsencrypt](https://letsencrypt.org/) project, that are working hard at encrypting all the things.
+ACME is part of the [Letsencrypt](https://letsencrypt.org/) project, which goal is to provide free SSL/TLS certificates with  automation of the acquiring and renewal process.
 
 ## Usage
 
 ```ruby
 # We're going to need a private key.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,24 +46,24 @@ challenge.filename # => ".well-known/acme-challenge/:some_token"
 # You can generate the body of the expected response.
 challenge.file_content # => 'string token and JWK thumbprint'
 
-# You can send no Content-Type at all but if you send one it has to be 'text/plain'.
+# You are not required to send a  Content-Type at all. If you do send one it has to be 'text/plain'.
```
jhass: "... it has to be the one returned from this method." ? Else this comment kinda makes the entire method redundant.

README.md
```diff
@@ -32,10 +35,10 @@ registration.agree_terms
 # We need to prove that we control the domain using one of the challenges method.
 authorization = client.authorize(domain: 'example.org')
 
-# For now the only challenge method supprted by the client is http-01.
+# For now the only challenge method supported by this client is http-01.
```
jhass: This is actually outdated.

louim: I did not dig into the code, can you tell me what other method are supported? Would it make sense to update the readme with example from other methods?

README.md
louim: You are right, that didn't make much sense! Fixed.

README.md
jhass: https://github.com/unixcharles/acme-client/blob/master/lib/acme/client/resources/authorization.rb#L19-L21

I'm not sure I'd add more examples here, it's probably time for some more structured documentation than a single big example.

unixcharles: We have full support for all methods implemented in Boulder, so I think we could just remove the mention.

The only missing challenge from the acme spec is proof-of-possession but I prefer not add it until a server side implementation is completed.

> I'm not sure I'd add more examples here, it's probably time for some more structured documentation than a single big example.
 
I agree, we could start linking to the wiki or create an example/ directory.

louim:
Removed the comment about the only type of challenge supported, rebased against current master and squashed. :+1:

unixcharles:
awesome! thanks @louim

Issue #50, opened 2016-01-20
Keep token for two steps validation


Soullivaneuh:
This issue is the following of #48 but with another issue.

I generate validation file and token on `acme_01.rb`, without validation:

```ruby
# We're going to need a private key.
require 'openssl'
private_key = OpenSSL::PKey::RSA.new(2048)

File.write('private_key', private_key)

print private_key

endpoint = 'https://acme-v01.api.letsencrypt.org/'

require 'acme/client'
client = Acme::Client.new(private_key: private_key, endpoint: endpoint)

registration = client.register(contact: 'mailto:soullivaneuh@gmail.com')
registration.agree_terms

authorization = client.authorize(domain: 'sullivansenechal.com')

challenge = authorization.http01

puts 'Filename: ' + challenge.filename # => ".well-known/acme-challenge/:some_token"
puts 'File content:' + challenge.file_content # => 'string token and JWK thumbprint'
puts 'Content type: ' + challenge.content_type
puts 'Token: ' + challenge.token

File.write('private_token', challenge.token)

FileUtils.mkdir_p( File.join( 'public', File.dirname( challenge.filename ) ) )
File.write( File.join( 'public', challenge.filename), challenge.file_content )
```

The problem is, when I want to launch `acme_02.rb` to validate the file, this one has changed.

I noticed that the challenge need a token, so I'm trying to store it and use it on the second script:

```ruby
# We're going to need a private key.
require 'openssl'
private_key = OpenSSL::PKey::RSA.new(File.read('private_key'))

print private_key

endpoint = 'https://acme-v01.api.letsencrypt.org/'

require 'acme/client'
client = Acme::Client.new(private_key: private_key, endpoint: endpoint)

authorization = client.authorize(domain: 'sullivansenechal.com')

challenge = authorization.http01(token: File.read('private_token'))

puts 'Filename: ' + challenge.filename # => ".well-known/acme-challenge/:some_token"
puts 'File content:' + challenge.file_content # => 'string token and JWK thumbprint'
puts 'Content type: ' + challenge.content_type
puts 'Token: ' + challenge.token

challenge.request_verification # => true
puts challenge.verify_status # => 'pending'
sleep(1)
puts challenge.verify_status # => 'valid'

# We're going to need a certificate signing request. If not explicitly
# specified, the first name listed becomes the common name.
csr = Acme::Client::CertificateRequest.new(names: %w[sullivansenechal.com www.sullivansenechal.com])

[...]

PR #53, opened 2016-02-01
Add ability to serialize challenges

itspriddle:
See #50 for background.

This allows one to store a challenge for later use. For example, the challenge can be created in 1 step and then verification can be completed separately by a background job.


unixcharles:
awesome :+1:

PR #54, opened 2016-02-02
Do not sign empty bodies, and set the Content-Type header on JWS signed requests

patch0:
As per section 5.1 of the ACME draft, only requests with a non-empty
body need signing.

This also stops the following warning from net/http

  Content-Type did not set; using application/x-www-form-urlencoded

when GET requests are sent with a JWS signed (but otherwise empty) body.


unixcharles:
Make sense. Do you have time to re-record the VCR cassettes?

patch0:
Hi, sorry, no I've not had time to update the VCR cassettes.  I'm not really familiar with VCR.

PR #55, opened 2016-02-06
Mark required Ruby version

will-in-wi:
In using the client, I noticed that it doesn't work on Ruby <2.1.0. I later realized that this is documented in the readme. This patch also adds this requirement to the gemspec.

Thanks for the gem! I'm using it in https://github.com/will-in-wi/letsencrypt-webfaction/. Works like a charm.


unixcharles:
:+1: thanks

PR #64, opened 2016-04-19
set Faraday request timeout

kintner:
@unixcharles What do you think of this? Should this be configurable? 

Currently the open / read timeout is 60 seconds if using [net/http](http://ruby-doc.org/stdlib-2.3.0/libdoc/net/http/rdoc/Net/HTTP.html#attribute-i-open_timeout). I need this to fail fast if LE is unresponsive or being slow.

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -9,8 +9,8 @@ class Acme::Client
     'revoke-cert' => '/acme/revoke-cert'
   }.freeze
 
-  def initialize(private_key:, endpoint: DEFAULT_ENDPOINT, directory_uri: nil)
-    @endpoint, @private_key, @directory_uri = endpoint, private_key, directory_uri
+  def initialize(private_key:, endpoint: DEFAULT_ENDPOINT, directory_uri: nil, connection_options: {})
+    @endpoint, @private_key, @directory_uri, @connection_options = endpoint, private_key, directory_uri, connection_options
     @nonces ||= []
     load_directory!
   end
@@ -62,7 +62,7 @@ class Acme::Client
   end
 
   def connection
-    @connection ||= Faraday.new(@endpoint) do |configuration|
+    @connection ||= Faraday.new(@endpoint, **@connection_options) do |configuration|
       configuration.use Acme::Client::FaradayMiddleware, client: self
       configuration.adapter Faraday.default_adapter
     end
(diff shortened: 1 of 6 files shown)


jhass:
I guess we should allow passing a custom faraday adapter and use it if our middleware is already present, else dup it and insert the middleware into it.

unixcharles:
I agree that 60 seconds is completely crazy

> I guess we should allow passing a custom faraday adapter or connection and use it if our middleware is already present, else dup it and insert the middleware into it.

Or maybe just the most common option that you might want config when initializing the client?

jhass:
Well deciding what's a common need is hard, Faraday already is an abstraction & unification layer, yet it has more options that I'd like to see passed through. A custom Faraday connection seems simpler to me.

unixcharles:
I just feel like injecting a middleware into an existing faraday connection is kind a weird.

In a case like this you might just want to create a new class that inherit from the client and re-implement it yourself.

What about doing something generic like:

```ruby
class Acme::Client
  def initialize(private_key:, endpoint: DEFAULT_ENDPOINT, directory_uri: nil, connection_options: {})
    @connection_options = connection_options
    # ...
  end

  def connection
    @connection ||= Faraday.new(@endpoint, **@connection_options) do |configuration|
      configuration.use Acme::Client::FaradayMiddleware, client: self
      configuration.adapter @connection_options.fetch(:adapter, Faraday.default_adapter)
    end
  end
end
```

jhass:
Well, I'd inject it into a copy, the injection itself isn't that weird IMO

```ruby
def prepend_middleware middleware
  return if @connection.builder.handlers.include? middleware

  @connection.builder.insert(0, middleware)
end
```

kintner:
@unixcharles How does this look now?

lib/acme/client.rb
```diff
@@ -62,7 +62,9 @@ def self.revoke_certificate(certificate, *arguments)
   end
 
   def connection
-    @connection ||= Faraday.new(@endpoint) do |configuration|
+    @connection ||= Faraday.new(@endpoint, **@connection_options) do |configuration|
+      configuration.options.open_timeout = 5
+      configuration.options.timeout = 5
```
jhass: This doesn't override the user supplied option?

kintner:
@jhass Yup! Thanks for catching!

unixcharles:
I like it 👍

Issue #62, opened 2016-04-12
Big warnings about circular require with ruby 2.3.0


lgromanowski:
Hi,
after upgrading ruby to v2.3.0 I see big warning message about circular require:

```
/home/lgromanowski/.rvm/gems/ruby-2.3.0/gems/json-jwt-1.5.2/lib/json/jose.rb:14: warning: method redefined; discarding old header
/home/lgromanowski/.rvm/gems/ruby-2.3.0/gems/json-jwt-1.5.2/lib/json/jws.rb:57: warning: method redefined; discarding old signature_base_string
/home/lgromanowski/.rvm/gems/ruby-2.3.0/gems/json-jwt-1.5.2/lib/json/jose.rb:14: warning: method redefined; discarding old header
/home/lgromanowski/.rvm/gems/ruby-2.3.0/gems/json-jwt-1.5.2/lib/json/jwe.rb:150: warning: method redefined; discarding old jwe_encrypted_key
/home/lgromanowski/.rvm/gems/ruby-2.3.0/gems/json-jwt-1.5.2/lib/json/jwe.rb:209: warning: method redefined; discarding old authentication_tag
/home/lgromanowski/.rvm/gems/ruby-2.3.0/gems/acme-client-0.3.1/lib/acme/client.rb:1: warning: loading in progress, circular require considered harmful - /home/lgromanowski/.rvm/gems/ruby-2.3.0/gems/acme-client-0.3.1/lib/acme-client.rb
        from /home/lgromanowski/.rvm/gems/ruby-2.3.0@global/gems/rake-11.1.2/lib/rake/rake_test_loader.rb:4:in  `<main>'
        from /home/lgromanowski/.rvm/gems/ruby-2.3.0@global/gems/rake-11.1.2/lib/rake/rake_test_loader.rb:4:in  `select'
        from /home/lgromanowski/.rvm/gems/ruby-2.3.0@global/gems/rake-11.1.2/lib/rake/rake_test_loader.rb:9:in  `block in <main>'
        from /home/lgromanowski/.rvm/gems/ruby-2.3.0@global/gems/rake-11.1.2/lib/rake/rake_test_loader.rb:9:in  `each'
        from /home/lgromanowski/.rvm/gems/ruby-2.3.0@global/gems/rake-11.1.2/lib/rake/rake_test_loader.rb:10:in  `block (2 levels) in <main>'
[...]

PR #66, opened 2016-04-29
Clean up gem loading requires

unixcharles:
Fix https://github.com/unixcharles/acme-client/issues/62

For some reason I wanted to be able to do both, `require 'acme-client'` and `require 'acme/client'`.

I can't recall a good reason for this. I guess just `require 'acme-client'` is fine.

I'll bump to `0.4` since this change the behaviour

cc @jhass @mac-adam-chaieb


mac-adam-chaieb:
👍

PR #70, opened 2016-06-16
Adds RateLimited error class.

nickmcdonnough:
"rateLimited" is an error code received when a client has hit a rate limit[0]. I am using this gem in a project and it would be nice to look for this specific error rather than having it thrown as the more generic Acme::Client::Error and then matching on the detail.

[0] https://ietf-wg-acme.github.io/acme/#errors


unixcharles:
I agree, ran into the same error too.

nickmcdonnough:
Thanks for the quick merge, Charles!

unixcharles:
included in the 0.3.2 release

PR #76, opened 2016-06-27
Make challenge_from_hash more strict with the arguments it receive

unixcharles:
I had `KeyError: key not found: :type` exception pop up in my stack down stream because I was passing unexpected `nil` to `#challenge_from_hash`.

Its not nice when generic error pop  from your dependencies.

I modified the method interface so that it return clear in errors message in those cases.

cc @maxboisvert @jhass

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -66,7 +66,12 @@ class Acme::Client
     end
   end
 
-  def challenge_from_hash(attributes)
+  def challenge_from_hash(arguments)
+    attributes = arguments.to_h
+    %w(type uri token).each do |key|
+      raise ArgumentError, "missing key: #{key}" unless attributes.key?(key)
+    end
+
     case attributes.fetch('type')
     when 'http-01'
       Acme::Client::Resources::Challenges::HTTP01.new(self, attributes)
@@ -74,6 +79,8 @@ class Acme::Client
       Acme::Client::Resources::Challenges::DNS01.new(self, attributes)
     when 'tls-sni-01'
       Acme::Client::Resources::Challenges::TLSSNI01.new(self, attributes)
+    else
+      raise 'Unsupported resource type'
     end
   end
 
--- a/spec/client_spec.rb
+++ b/spec/client_spec.rb
@@ -206,9 +206,9 @@ describe Acme::Client do
     end
 
     it 'returns nil if an unsupported challenge type is provided' do
-      challenge = client.challenge_from_hash(challenge_hash.merge('type' => 'nope'))
-
-      expect(challenge).to be_nil
+      expect {
+        client.challenge_from_hash(challenge_hash.merge('type' => 'nope'))
+      }.to raise_error(RuntimeError, 'Unsupported resource type')
     end
   end
 end


jhass:
So the call side interface is supposed to do `**JSON.parse(challenge_file)` etc.? I guess that's fine with me.

unixcharles:
It will change the interface in a way the require symbols so that might be a problem

jhass:
Yeah, I guess it would be more like manually unpacking the serialized challenge. Perhaps manually verifying the keyset and raising an `ArgumentError` is nicer?

maxboisvert:
I like this change. Another way of doing it could be to keep the attributes parameter and replace `attributes.fetch('type')` by `attributes.['type']`. `Unsupported resource type' ` would then be raised.

unixcharles:
updated the PR to keep the hold string key interface

lib/acme/client.rb
```diff
@@ -66,14 +66,19 @@ def connection
     end
   end
 
-  def challenge_from_hash(attributes)
-    case attributes.fetch('type')
+  def challenge_from_hash(arguments)
+    arguments.default_proc = lambda { |_, key| raise "Missing #{key} key" }
```
jhass: Sorry I don't like this much at all:

* It's raising a `RuntimeError`, if one would want to handle it, one would need to interpret the associated message. This arguably worse than the standard `KeyError`.
* `Hash#default_proc=` is modifying the object, so this modifies the callers object and can have unexpected and surprising side effects in case the caller still uses the passed value elsewhere afterwards.

unixcharles: > It's raising a RuntimeError, if one would want to handle it, one would need to interpret the associated message. This arguably worse than the standard KeyError.

I'm kind of okay with this as this is intended to be a message for the developers using the gem. The goal being communicating that they are not correctly using the interface. I don't see a case were you would want to rescue this error for control flow or anything. Do you see a use case for that?

> Hash#default_proc= is modifying the object, so this modifies the callers object and can have unexpected and surprising side effects in case the caller still uses the passed value elsewhere afterwards.

Good point. `Hash#dup` ?

maxboisvert: Why not just add a guard check like this :

`%w(type uri token)`.each { |key| raise "Something" unless attributes.key?(key) }

jhass: Well I guess, a bit contrived example would be that they have some old hashes after we added another key to the API for some reason and just want to log an error in that case and move on, but still escalate other issues higher.

I think using `Hash#default_proc=` is simply trying to be too smart here tbh, something like this should be good enough:
```rb
def challenge_from_hash(attributes)
  type  = attributes.fetch(:type)
  uri   = attributes.fetch(:uri)
  token = attributes.fetch(:token)

  # ...
end
```
It won't change the error for the initial motivation of the PR but I think we came to the conclusion that there's not much to do about that. If you do want to change the error message nonetheless, one could add `, &KEY_ERROR_PROC)` to each.

unixcharles:
```ruby
    arguments = arguments.to_h
    %w(type uri token).each do |key|
      raise ArgumentError, "missing key: #{key}" unless attributes.key?(key)
    end
```

?

maxboisvert:
👍

PR #79, opened 2016-07-04
Fix error handling on non json response

unixcharles:
There was a service disruption today and LetsEncrypt started sending non JSON body errors.

ping @mlhamel @jhass


jhass:
So what's the behavior now? I think this would greatly benefit from some tests :)

unixcharles:
Atm it raise because it doesn't expect the body to be anything but a hash. I'll add some test

jhass:
Yeah I got that much, I wonder it should still raise, no? What's the new error?

unixcharles:
It will raise the generic `Acme::Client:Error` from the caller if there is no error name and it will include the body as part of the error message.

Gemfile
```diff
@@ -5,4 +5,8 @@ group :development, :test do
   gem 'pry'
   gem 'rubocop', '0.36.0'
   gem 'ruby-prof', require: false
+
+  if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new('2.2.2')
```
mlhamel: arrgg but if we have no choice 👍

PR #83, opened 2016-07-15
allow to require gem as: require 'acme/client'

porbas:
it's kind of standard according to http://guides.rubygems.org/name-your-gem/

### DIFF
--- a/lib/acme-client.rb
+++ b/lib/acme-client.rb
@@ -1,5 +1,5 @@
 # This file lives here for backward compatibility only and should be deprecated soon
-# gem named acme-client should be required as 
+# gem named acme-client should be required as
 # require 'acme/client'
 
 require 'faraday'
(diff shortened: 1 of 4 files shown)


lib/acme-client.rb
```diff
@@ -1,3 +1,7 @@
+# This file lives here for backward compatibility only and should be deprecated soon
+# gem named acme-client should be required as
```
jhass: It should stay forever too, iirc for example bundler autorequire will still just try plain gem name.

porbas: Bundler "knows" about namespaced libs and would autorequire acme/client without complaints if there wouldn't be acme-client.rb file.
https://github.com/bundler/bundler/blob/master/lib/bundler/runtime.rb#L97

jhass: TIL, thanks. I still think it can't harm and it seems to be common practice to offer both.

unixcharles:
👍

PR #87, opened 2016-07-16
Add user agent to calls made to upstream. Fixes #86

### LINKED ISSUE #86: Set Faraday User Agent for requests
Let's Encrypt comments in a ticket (https://github.com/letsencrypt/boulder/issues/1902#issuecomment-233032593) that they are looking at logs to see how people use them and to identify clients when they might break something.

Right now, it looks like Acme::Client uses the default Faraday header, which probably isn't especially helpful. Could we give them a hand and send an application specific header?

Thanks!

will-in-wi:
Fixes https://github.com/unixcharles/acme-client/issues/86

### DIFF
--- a/spec/client_spec.rb
+++ b/spec/client_spec.rb
@@ -30,7 +30,7 @@ describe Acme::Client do
 
     before(:each) do
       stub_request(:head, 'http://127.0.0.1:4000/acme/new-reg').with(
-        headers: { 'Accept' => '*/*', 'User-Agent' => 'Acme::Client v0.3.6' }
+        headers: { 'Accept' => '*/*', 'User-Agent' => 'Acme::Client v0.3.6 (https://github.com/unixcharles/acme-client)' }
       ).to_timeout
     end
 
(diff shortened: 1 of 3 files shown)


jhass:
How about adding the URL to this repo?

will-in-wi:
URL added

spec/client_spec.rb
```diff
@@ -30,7 +30,7 @@
 
     before(:each) do
       stub_request(:head, 'http://127.0.0.1:4000/acme/new-reg').with(
-        headers: { 'Accept' => '*/*', 'User-Agent' => 'Faraday v0.9.2' }
+        headers: { 'Accept' => '*/*', 'User-Agent' => 'Acme::Client v0.3.6 (https://github.com/unixcharles/acme-client)' }
```
jhass: I don't think we should hardcode the version here :)

will-in-wi: Ah, good point. I had fixed that in an initial draft and then hard coded to simplify debugging. I forgot to set it back. Fixed.

unixcharles:
merged in https://github.com/unixcharles/acme-client/pull/89

I fixed the rubocop failure and moved the agent in a constant.

PR #95, opened 2016-08-10
Require time stdlib

jhass:
https://github.com/unixcharles/acme-client/blob/92debc00c5736813f97a6012ca7fdefeee765236/lib/acme/client/resources/authorization.rb#L28-L28

`Time.iso8601` is only available after requiring the standard library. Most client apps will do this somewhere explicitly or implicitly through one of  their other dependencies, but since we make use of it we should do declare it explicitly.

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -6,6 +6,7 @@ require 'openssl'
 require 'digest'
 require 'forwardable'
 require 'base64'
+require 'time'
 
 module Acme; end
 class Acme::Client; end


unixcharles:
good catch

will-in-wi:
Could this get released shortly? This is causing new installations to fail: https://github.com/will-in-wi/letsencrypt-webfaction/issues/39

Thanks!

unixcharles:
just released 0.4.1

will-in-wi:
Thanks!

Issue #97, opened 2016-08-12
Set X509 version when generating self-signed certificates


rolandshoemaker:
Follow up to #84, this also needs to be done when using OpenSSL to generate self-signed certificates for `tls-sni` type challenges.

phillipp:
Would it be possible to release with this so the client will continue working after Let's Encrypt deploys the new version (anytime soon)?

unixcharles:
Just released 0.4.1

phillipp:
Thank you!

PR #98, opened 2016-08-12
Set the X509 version of the self-signed certificate

phillipp:
This is a PR to fix https://github.com/unixcharles/acme-client/issues/97

Original boulder issue was https://github.com/letsencrypt/boulder/issues/2124


unixcharles:
thanks!

PR #99, opened 2016-08-15
Add Authorization#verify_status

jhass:
Recent versions of boulder will no longer process a challenge if the
associated authorization is already valid, that is another challenge was
previously solved. This means we need to allow people to poll on the
authorizations status rather than the challenge status so they don't
have to poll on the status of all challenges of an authorization all the
time.

See
https://community.letsencrypt.org/t/upcoming-change-valid-authz-reuse/16982
and https://github.com/letsencrypt/boulder/issues/2057


README.md
```diff
@@ -90,11 +94,18 @@ challenge = client.challenge_from_hash(JSON.parse(File.read('challenge')))
...
 
 # Wait a bit for the server to make the request, or just blink. It should be fast.
 sleep(1)
 
+# Rely on authorization.verify_status more than on challenge.verify_status,
+# if the former is 'valid' you can already issue a certificate and the status of
+# the challenge is not relevant and in fact may never change from pending.
+# However if authorization.verify_status is 'invalid', you can get at the error
+# message only through the challenge.
+authorization.verify_status # => 'valid'
```
unixcharles: maybe we should just not mention that you can verify the status from the challenge?

lib/acme/client/resources/authorization.rb
```diff
@@ -7,12 +7,26 @@ class Acme::Client::Resources::Authorization
...
+    @uri = response.headers['Location']
     assign_attributes(response.body)
   end
 
+  def verify_status
+    response = @client.connection.get(@uri)
+
+    assign_attributes(response.body)
+    status
+  end
```
unixcharles: I assume some users will need to be able to serialize / load this object now that you can query on it?

README.md
jhass: Well, when `client.authorize(...).status` isn't `valid` you're going to submit a challenge and that challenge will be processed because the authorization wasn't `valid`, so for some code designs not keeping the authorization around but just the challenge can work and be valid usage.

lib/acme/client/resources/authorization.rb
jhass: Mh, maybe. How about exposing `uri` and having a `Client#fetch_authorization(uri)`?

jhass: We could do something similar for the challenges in fact.

unixcharles: > We could do something similar for the challenges in fact.

👍

jhass: Do you want me to do it here or is it okay as a follow up?

README.md
unixcharles: I just mean for the demo in the README, not the actual implementation. It might be less confusing.

jhass: Well, the error message won't be populated in the existing challenge (but in the new `authorization.challenge_type`) until `verify_status` is called on it.

lib/acme/client/resources/authorization.rb
unixcharles: I think it would make sense, this way I can release a new gem once I merge.

jhass:
Okay, while working on I came to the opinion that we should go the full step. Now the challenge endpoints are never polled, `Challenge#verify_status` simply delegates to `Authorization#verify_status`. The latter updates all challenge objects.

Since constructing a challenge without an authorization then no longer is leading to a consistent state, I dropped `Client#challenge_from_hash` and `Challenge#to_h` in favor of `Client#fetch_authorization` and `Authorizatio#uri` respectively. This is the main backwards incompatible change that will require users who stored challenges to drain their queues before updating.

lib/acme/client.rb
```diff
@@ -88,24 +93,6 @@ def connection
...
-      Acme::Client::Resources::Challenges::HTTP01.new(self, attributes)
-    when 'dns-01'
-      Acme::Client::Resources::Challenges::DNS01.new(self, attributes)
-    when 'tls-sni-01'
-      Acme::Client::Resources::Challenges::TLSSNI01.new(self, attributes)
-    else
-      raise 'Unsupported resource type'
-    end
-  end
-
```
unixcharles: I think we should bump to 1.0 if we're doing a major change like this.

jhass: Well, technically SemVer allows breaking API changes on minor releases while < 1.0, but it's up to you. I just don't see a sane way to support both APIs.

jhass: But perhaps time for a changelog where we can announce it.

unixcharles: Just thinking it might be a good time to go for 1.0 and start taking API stability more into consideration.

The gem is pretty stable at this point I would say. I already provisioned a bit more than 1,080,000 certificates with it.

unixcharles:
My only concern with this is that If we remove the ability to serialize / deserialize challenges, how do we call `#request_verification` at a later time?

jhass:
`client.fetch_authorization(stored_uri).http01.request_verification`.

unixcharles:
> `client.fetch_authorization(stored_uri).http01.request_verification`.

Oh right, nevermind lol

whitehat101:
This makes sense.

I was considering storing challenges, but they don't include the domain or the expiration (which is what I really want to query offline). 

As far as I can tell, for a given domain`@client.authorize(domain: domain)` always returns the same URL in the location header. The only difference between `authorize` and `fetch_authorization` is one uses 1 POST and other other uses 1 GET.

Why would one prefer `fetch_authorization` over `authorize`?

It seems like I could simply track the domain, and there isn't much benefit to tracking the URI directly. Maybe you could verify_status on that without a client, but to `request_verification` or do most useful things, you'd still need a client.

Since the `authorize` method already takes keyword args, why not fold `fetch_authorization` into it? 

```ruby
@client.authorize(domain: domain).http01.request_verification
@client.authorize(uri: uri).http01.request_verification
```

PR #100, opened 2016-09-21
Allow query the endpoint of the client object

jorhett:
It would be awful handy to ask the client object which endpoint it was configured with when reporting errors, writing logs, etc.


unixcharles:
thanks

PR #106, opened 2016-11-23
Set serial number of self-signed certificate

phillipp:
Following up on #97 the serial number must be set, too.

This pull request adds a static serial number.


phillipp:
@unixcharles Is this going to be  merged (and a new version released) soon? I still can't use your version but need to use my fork because TLS SNI validation doesn't work without the patch since November 2016...

unixcharles:
Sorry about that.

phillipp:
Thanks bro

Issue #108, opened 2016-12-27
Undefined challenge_from_hash but it's in read me 


ArturT:
In the readme you can find example:

```ruby
# Load a saved challenge. This is only required if you need to reuse a saved challenge as outlined above.
challenge = client.challenge_from_hash(JSON.parse(File.read('challenge')))
```

but the method `challenge_from_hash` doesn't exist. I found it was removed from code base:
https://github.com/unixcharles/acme-client/compare/3d41fa27cdc80e61db6685c3829447c9aa49fd70...9a80e9301acd3f17ebff6ceaf245e0c400d08519
This is the diff between version 0.3.1 and 0.5.0.

Is there a reason why it was removed? What is recommended way to create challenge from file then?

jhass:
Please see the discussion in #99

PR #110, opened 2016-12-27
Add example for fetch_authorization instead of removed challenge_from_hash in README

ArturT:
Related issue: 
https://github.com/unixcharles/acme-client/issues/108

Related PR:
https://github.com/unixcharles/acme-client/pull/99


unixcharles:
Thanks @ArturT

PR #111, opened 2016-12-28
Enable ECDSA certificates

sdaubert:
Let's Encrypt supports signing ECDSA certificates since February 2016.

To enable acme-client supports ECDSA certificate requests, a minor change is needed in `CertificateRequest#generate`: set private key instead of public key into `csr.public_key` to enable ECDSA certificates.


lib/acme/client/certificate_request.rb
```diff
@@ -79,7 +79,7 @@ def validate_subject_common_name
 
   def generate
     OpenSSL::X509::Request.new.tap do |csr|
-      csr.public_key = @private_key.public_key
```
unixcharles: This is a pretty weird change to be honest. Doesn't `OpenSSL::PKey::EC` respond to public key?

I understand that the private key would work here but I'm concerned to passing down the stack a private key when its not really needed.

unixcharles:
I've tried to write a test locally for this but I'm running into some issues.

Not calling `OpenSSL::PKey::EC#generate_key` and trying to generate a CSR will segfault ruby. see https://travis-ci.org/unixcharles/acme-client/builds/187947222)

Once you call `generate_key` on the instance you get to `NoMethodError: undefined method `private?'`. My conclusion is that for some reason `OpenSSL::PKey::EC` is not respecting the compatibility interface with the other `OpenSSL::PKey`.

Does changing that line fix it for you? Did you someone monkey patch `OpenSSL::PKey::EC` to get it working?

lib/acme/client/certificate_request.rb
sdaubert: `OpenSSL::PKey::EC` is a private key type. Unfortunaltely, OpenSSL::X509::Request does not accept EC public key (`OpenSSL::PKey::EC::Point` class) so we have to use private key here.
This is a known limitation for EC use with ruby.

sdaubert:
Yes, I did some monkey patches as OpenSSL::PKey::EC is not respecting the compatibility interface with the otherOpenSSL::PKey. This is fixed in ruby 2.4.0.

Classic monkey patch is:
```ruby
class OpenSSL::PKey::EC
  alias private? private_key?
  alias public? public_key?
end
```

sdaubert:
What do you plan for this PR?
Do you want me to do some updates?

unixcharles:
@sdaubert sorry for the silence on this. I'm a bit concerned of monkey patching the standard library.

Do you think you could subclass `OpenSSL::PKey::EC` into a `Acme::Client::CertificateRequest::ECPKeyPatch` class or something and add the compatibility patch for `OpenSSL::X509::Request` there?

sdaubert:
@unixcharles This seems possible. I will have a look at it.

lib/acme/client/certificate_request.rb
```diff
@@ -79,7 +79,15 @@ def validate_subject_common_name
...
+      if @private_key.is_a?(OpenSSL::PKey::EC) && RbConfig::CONFIG['MAJOR'] == '2' &&
+         RbConfig::CONFIG['MINOR'].to_i < 4
+        # OpenSSL::PKey::EC does not respect classic PKey interface (as defined by
+        # PKey::RSA and PKey::DSA) until ruby 2.4.
+        # Supporting this interface needs monkey patching of OpenSSL:PKey::EC, or
+        # subclassing it. Here, use a subclass.
+        @private_key = ECKeyPatch.new(@private_key)
+      end
+      csr.public_key = @private_key
```
unixcharles: Can this go in `ECKeyPatch` as well?

sdaubert: What do you want to go in `ECKeyPatch`? last line? or using private key instead of public one?

unixcharles: What's the reason for not doing `csr.public_key = @private_key.public_key` ?

The `EC` doesn't respond to `public_key`?

sdaubert: EC responds to `public_key`, but OpenSSL::X509::Request cannot handle its return, a `OpenSSL::PKey::EC::Point`. So, we have to use private key, from which public key is retrieve by openssl library.

sdaubert:
thanks!

PR #113, opened 2017-02-25
Update Travis Rubies

bdewater:
Ruby 2.4 ships with OpenSSL gem 2.0, make sure everything still works.


bdewater:
Something's weird going on 🤔 and I've hidden it behind a test helper...

Ruby 2.3.1:
```
request = Acme::Client::CertificateRequest.new(names: %w(example.org www.example.org), private_key: test_key)
request.csr.attributes.first.value.first.first.value.first.value
=> "subjectAltName"
```
Ruby 2.4.0:
```
request = Acme::Client::CertificateRequest.new(names: %w(example.org www.example.org), private_key: test_key)
request.csr.attributes.first.value.value.first.value.first.value.first.value
=> "subjectAltName"
```

spec/certificate_request_spec.rb
```diff
@@ -130,10 +130,10 @@
     request = Acme::Client::CertificateRequest.new(names: %w(example.org www.example.org), private_key: test_key)
 
     extension = request.csr.attributes.find { |attribute|
-      attribute.value.first.first.value.first.value == 'subjectAltName'
```
unixcharles: Such a weird API change. Do you happen to know what lead to that change? I couldn't find anything in the openssl change log / repo.

spec/support/asn1_helper.rb
```diff
@@ -0,0 +1,11 @@
+module Asn1Helper
+  def asn1_dig(obj)
+    if obj.respond_to?(:first) && obj.first.is_a?(Enumerable)
+      asn1_dig obj.first
+    elsif obj.respond_to?(:value) && obj.value.is_a?(Enumerable)
+      asn1_dig obj.value
+    else
+      obj
+    end
```
unixcharles: I wonder if we shouldn't just specifically do conditional assertion on `OpenSSL::VERSION`? What do you think?

bdewater: Yeah, it's definitely more clear but might break again somewhere in the future. Then again, ASN1 is always a PITA to work with so let's deal with it when the problem arises 😀

unixcharles: I would personally prefer it to fail. I'm scare this could unexpectedly work, which would be worst 😄

unixcharles:
Thanks Bart!

PR #116, opened 2017-03-06
Fix error_name in faraday_middleware.rb

### LINKED ISSUE #115: Two-word errors will never be instantiated
Due to the way `error_name` works, two word errors won't be instantiated when talking to ACME servers conforming to [the spec](https://letsencrypt.github.io/acme-spec/#rfc.section.5.4).

error names look like: `badNonce`, but `error_name` in `lib/acme/client/faraday_middleware.rb` uses `capitalize`, which capitalizes the first character of a string and lowercase the rest - resulting in `error_class` looking for `Badnonce`, not the correct `BadNonce`

I can submit a PR with a fix for that, but I don't have a good idea for how to test for it with your test setup.

telyn:
Due to the way error_name works, two word errors won't be instantiated when talking to ACME servers conforming to the spec.

This PR corrects the instantiation of errors issue I reported in #115 by modifying `error_name` to upcase the first character of error types and leave the rest as-is - `String.capitalize` upcases the first character and downcases the rest.

Fixes #115

### DIFF
--- a/lib/acme/client/faraday_middleware.rb
+++ b/lib/acme/client/faraday_middleware.rb
@@ -69,7 +69,7 @@ class Acme::Client::FaradayMiddleware < Faraday::Middleware
   end
 
   def error_type_to_klass(type)
-    type.gsub('urn:acme:error:', '').split(/[_-]/).map { |type_part| type_part[0].upcase + type_part[1..-1] }.join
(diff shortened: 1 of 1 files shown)


unixcharles:
Good catch, thanks!

PR #120, opened 2017-05-16
Support external account keys

### LINKED ISSUE #117: Error when register with an ECDSA account key
I am trying to register with an ECDSA key.
I modified `Acme::Client::Crypto#initialize` to avoid problems encountered in #111:
```ruby
  def initialize(private_key)
    if private_key.is_a?(OpenSSL::PKey::EC) && RbConfig::CONFIG['MAJOR'] == '2' &&
       RbConfig::CONFIG['MINOR'].to_i < 4
      @private_key = Acme::Client::CertificateRequest::ECKeyPatch.new(private_key)
    else
      @private_key = private_key
    end
  end
```

In my code, when calling `client.register`, I have this exception:
```
Acme::Client::Error::Malformed: JWS verification error
         # ./lib/acme/client/faraday_middleware.rb:43:in `raise_on_error!'
         # ./lib/acme/client/faraday_middleware.rb:33:in `on_complete'
         # ./lib/acme/client/faraday_middleware.rb:18:in `block in call'
         # ./vendor/bundle/ruby/2.3.0/gems/faraday-0.10.0/lib/faraday/response.rb:61:in `on_complete'
         # ./lib/acme/client/faraday_middleware.rb:18:in `call'
         # ./vendor/bundle/ruby/2.3.0/gems/faraday-0.10.0/lib/faraday/rack_builder.rb:139:in `build_response'
         # ./vendor/bundle/ruby/2.3.0/gems/faraday-0.10.0/lib/faraday/connection.rb:377:in `run_request'
         # ./vendor/bundle/ruby/2.3.0/gems/faraday-0.10.0/lib/faraday/connection.rb:177:in `post'
         # ./lib/acme/client.rb:45:in `register'
```

I may provide a PR with updated Crypto code and a spec on Acme::Client#register, if you need it.

mastahyeti:
I am wanting to use an account key that exists on a remote system (like an HSM). This PR abstracts the `Crypto` class to work with any object conforming to a `JWK::Base` "abstract class". This allows users of this library to implement their own `JWK` classes that can be used by the `Crypto` class.

I went to great lengths not to break the `Crypto` API, since I'm assuming this library does semantic versioning and probably doesn't want to cut a new major version. If that's not such an issue, I think some of these APIs could be cleaned up with the addition of the `JWK` classes.

I think this PR also fixes #117.

/cc @parkr


mastahyeti:
I guess this still needs some work for it to be usable. We'll have to have a `jwk` accessor on the `client` class and have the Faraday middleware use that instead of `Client#private_key`

unixcharles:
Pretty good stuff @mastahyeti for far, I think its a good direction for the library.

As for the `Crypto` class, I would not consider it has a public class, so I don't think it would be a bit deal to change its interface and bump to `0.6.0`.

Let me know when you are done and I'll give it a detailed review.

unixcharles:
cc @davidcornu

mastahyeti:
I moved the JWS/thumbprint logic into the `JWK::Base` class, allowing me to get rid of the `Crytpo` class altogether. I also wired things up so you can specify a JWK *or* a private key when instantiating a `Client`. This should make the library useable with custom JWK implementations. I think this PR is ready for some :eyes:.

parkr:
This looks great to me! I'd double check that we can remove the methods removed in this PR (for backwards compatibility) but otherwise 👍 :shipit: ✨

mastahyeti:
Gentle bump: have you had a chance to give this diff a look @unixcharles?

spec/support/ssl_helper.rb
```diff
@@ -12,16 +12,41 @@ def initialize
...
+        generate_ecdsa_key('secp384r1')
+      when 3
+        generate_ecdsa_key('secp521r1')
+      end
+    end
+
+    def generate_ecdsa_key(curve)
+      k = OpenSSL::PKey::EC.new(curve)
+      k.generate_key
+      Acme::Client::CertificateRequest::ECKeyPatch.new(k)
```
unixcharles: Its more of an internal object, so you can use `OpenSSL::PKey::EC` normally here.

`ECKeyPatch` is only there to avoid monkey patching openssl. Its because `OpenSSL::PKey::EC` does not respect the interface of `OpenSSL::PKey::RSA` and it break compatibility with the `OpenSSL::X509::Request`.

spec/support/ssl_helper.rb
```diff
@@ -12,16 +12,41 @@ def initialize
...
     private
 
     def load
       if File.exist?(KEYSTASH_PATH)
-        YAML.load_file(KEYSTASH_PATH).map { |key| OpenSSL::PKey::RSA.new(key) }
+        YAML.load_file(KEYSTASH_PATH).shuffle.map do |pem|
+          begin
+            OpenSSL::PKey::RSA.new(pem)
+          rescue
+            Acme::Client::CertificateRequest::ECKeyPatch.new(pem)
```
unixcharles: same here

spec/support/ssl_helper.rb
mastahyeti: Plain `OpenSSL::PKey::EC` keys are pretty much unusable because they don't define `#private?`. [Here's](https://gist.github.com/mastahyeti/3fbb84a9fc482a0fae11c8bacc9d4d3a) the test output if you try.

unixcharles: Oh I see. It worked because its fixed in the latest ruby releases.

I tried again with `2.1` and I got the same issue as you.

Shouldn't the client wrap the `OpenSSL::PKey::EC` with `ECKeyPatch` whenever its needed then?

It just feel strange to me that we expose the developers to this. Especially knowing that this patch will not be required in future ruby.

mastahyeti: I've addressed this in other projects by defining singleton methods on the keys. We could add a util method like

```ruby
def fix_key(key)
  return key if key.respond_to?(:private?) && key.respond_to?(:public?)

  key.define_singleton_method(:private?) { private_key? }
  key.define_singleton_method(:public?) { public_key? }

  key
end
```

And call it wherever we're passed a key.

mastahyeti:
I tried this branch out against the Let's Encrypt staging endpoint and I'm getting an exception:

```
irb(main):005:0* GitHub.acme.register(contact: "mastahyeti@gmail.com")
Acme::Client::Error::Malformed: JWS verification error
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/acme-client-0.5.5.8.g9d6be13/lib/acme/client/faraday_middleware.rb:43:in `raise_on_error!'
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/acme-client-0.5.5.8.g9d6be13/lib/acme/client/faraday_middleware.rb:33:in `on_complete'
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/acme-client-0.5.5.8.g9d6be13/lib/acme/client/faraday_middleware.rb:18:in `block in call'
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/faraday-0.9.2/lib/faraday/response.rb:57:in `on_complete'
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/acme-client-0.5.5.8.g9d6be13/lib/acme/client/faraday_middleware.rb:18:in `call'
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/faraday-0.9.2/lib/faraday/rack_builder.rb:139:in `build_response'
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/faraday-0.9.2/lib/faraday/connection.rb:377:in `run_request'
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/faraday-0.9.2/lib/faraday/connection.rb:177:in `post'
	from /Users/mastahyeti/Projects/github/vendor/gems/2.4.0/ruby/2.4.0/gems/acme-client-0.5.5.8.g9d6be13/lib/acme/client.rb:56:in `register'
	from (irb):5
	from script/console:76:in `<main>'
```

This is the same error as from https://github.com/unixcharles/acme-client/issues/117, so it sounds like some more debugging is required. I'll try to dig in.

mastahyeti:
Looks like the JOSE code in Boulder isn't expecting the signature to be ASN.1 encoded. I'll push up a fix.

unixcharles:
thanks for the heads up

mastahyeti:
I fixed the signature encoding issue in 2c25a0b. Any further thoughts on how we should work around the broken `OpneSSL::PKey::EC` API?

unixcharles:
> I fixed the signature encoding issue in 2c25a0b. Any further thoughts on how we should work around the broken OpneSSL::PKey::EC API?

I think we're good to merge as is. I'll think of something for the ec-patch thing for the next release.

@davidcornu you mind giving it a 👀 ?

lib/acme/client/faraday_middleware.rb
```diff
@@ -14,7 +14,7 @@ def initialize(app, client:)
   def call(env)
     @env = env
     @env[:request_headers]['User-Agent'] = USER_AGENT
-    @env.body = crypto.generate_signed_jws(header: { nonce: pop_nonce }, payload: env.body)
+    @env.body = client.jwk.jws(header: { nonce: pop_nonce }, payload: env.body)
```
davidcornu: Is `env.body` a string or is setting the body to a serializable object the convention for Faraday?

lib/acme/client/jwk/base.rb
```diff
@@ -0,0 +1,77 @@
...
+  # Returns a JSON string.
+  def to_json
+    to_h.to_json
+  end
+
+  # Get this JWK as a Hash for JSON serialization.
+  #
+  # Returns a Hash.
+  def to_h
+    raise 'not implemented'
```
davidcornu: `raise NotImplementedError`

lib/acme/client/jwk/base.rb
```diff
@@ -0,0 +1,77 @@
...
+      alg: jwa_alg,
+      jwk: to_h
+    }
+  end
+
+  # The name of the algorithm as needed for the `alg` member of a JWS object.
+  #
+  # Returns a String.
+  def jwa_alg
+    raise 'not implemented'
```
davidcornu: `raise NotImplementedError`

lib/acme/client/jwk/base.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  # Sign a message with the private key.
+  #
+  # message - A String message to sign.
+  #
+  # Returns a String signature.
+  # rubocop:disable Lint/UnusedMethodArgument
+  def sign(message)
+    raise 'not implemented'
```
davidcornu: `raise NotImplementedError`

lib/acme/client/jwk/ecdsa.rb
```diff
@@ -0,0 +1,113 @@
...
+    'secp521r1'  => 'P-521'
+  }.freeze
+
+  # Mapping from OpenSSL curve names to associated JWS "alg" names.
+  # https://tools.ietf.org/html/rfc7518#section-3.1
+  JWA_ALG_MAP = {
+    'prime256v1' => 'ES256',
+    'secp384r1'  => 'ES384',
+    'secp521r1'  => 'ES512'
+  }.freeze
```
davidcornu: nit - Why not combine these four constants into a single hash?

```ruby
KNOWN_CURVES = {
  'prime256v1' => {
    jwa_crv: 'P-256',
    jwa_alg: 'ES384',
    digest: OpenSSL::Digest::SHA256
  },
  'secp384r1' => {
    jwa_crv: 'P-384',
    jwa_alg: 'ES384',
    digest: OpenSSL::Digest::SHA384
  },
  'secp521r1' => {
    jwa_crv: 'P-521',
    jwa_alg: 'ES512',
    digest: OpenSSL::Digest::SHA512
  }
}.freeze

```

lib/acme/client/jwk/base.rb
```diff
@@ -0,0 +1,77 @@
+class Acme::Client::JWK::Base
```
davidcornu: Seems odd to me to have a base class not define the initialization signature

lib/acme/client/util.rb
```diff
@@ -0,0 +1,59 @@
+module Acme::Client::Util
+  def urlsafe_base64(data)
+    Base64.urlsafe_encode64(data).sub(/[\s=]*\z/, '')
+  end
+
+  # Sets public key on CSR or cert.
```
davidcornu: Are the methods below used outside of tests? If not I'd put them in a module in `spec/support`

lib/acme/client/faraday_middleware.rb
unixcharles: Its a string but a middleware could misbehave and change it for something else. Its generally assume that its going to be a string.

mastahyeti: Yeah. It looks like the built-in Faraday adapters are all happy with anything responding to `#to_s`.

lib/acme/client/jwk/base.rb
mastahyeti: Any initialization args are implementation-dependent. Should I add something like this?

```ruby
def initialize
  raise NotImplementedError
end
```

davidcornu: @mastahyeti maybe

```ruby
def initialize(_private_key)
  raise NotImplementedError
end
```

mastahyeti: That matches the subclasses defined in this repo, but isn't very general. For example, the JWK that I'm using in my project has a signature like

```ruby
class HsmBackedJwk < Acme::Client::JWK::Base
  def initialize(hsm_key_label)
    ...
  end
end
```

I don't feel strongly though since subclasses will be overriding the method anyway. I'm happy to add it with the `_private_key` argument.

davidcornu: > That matches the subclasses defined in this repo, but isn't very general.

Ah fair enough. I assumed based on the ones implemented that callers could expect a similar initialization from subclasses of `Acme::Client::JWK::Base` but if it isn't generalisable then you're probably right to leave it out.

mastahyeti:
Thanks for the :eyes: folks. I think I've addressed everyone's feedback in the last couple commits. Let me know if I'm missing anything.

sdaubert:
I am sorry but i have encountered an issue with this PR.

When using a 256-bit private key, Acme::Client#register raises:
```
Acme::Client::Error::Malformed:
       signature type 'ES384' in JWS header is not supported, expected one of RS256, ES256, ES384 or ES512
```
With a 256-bit private key, algorithm name should be ES256, and not ES384. This is resolved by updating `Acme::Client::JWK::ECDSA::KNWON_CURVES['prime256v1'][:jwa_alg]` to `'ES256'`.

mastahyeti:
Yeah. I just [noticed](https://github.com/unixcharles/acme-client/pull/120#commitcomment-22368360) this too. I'll open a PR.

sdaubert:
@mastahyeti I did not refresh page before posting, sorry :confused:

PR #127, opened 2017-07-11
Rate limited retry after

jeromegn:
I just hit a rate limited error when creating a new certificate (due to a bug, that created a loop of trying to create a certificate)

I thought I'd add this `retry_after` property to `Acme::Client::Error::RateLimited` instances (when available.)

According to the spec, the server may return a `Retry-After` header containing a value in seconds after which one may retry the same operation.

Let me know if you'd like this done another way.


lib/acme/client/error/rate_limited.rb
```diff
@@ -0,0 +1,7 @@
+class Acme::Client::Error::RateLimited < Acme::Client::Error
+  DEFAULT_RETRY_AFTER = 10
+
+  def retry_after
+    @env.response_headers.fetch('Retry-After', DEFAULT_RETRY_AFTER).to_i
+  end
+end
```
unixcharles: CI is being picky about the newline

jeromegn: Fixed!

cpu:
> According to the spec, the server may return a Retry-After header containing a value in seconds after which one may retry the same operation.

That's true, but I wanted to point out that if you're targetting Let's Encrypt and the Boulder implementation of the ACME spec we don't [presently send this Retry-After header](https://github.com/letsencrypt/boulder/blob/9c7482fa9465cb77fe7da90175bfaab599e9af6c/docs/acme-divergences.md#section-65).

jeromegn:
@cpu oh, derp. Thanks for pointing that out. Was hoping to be rescue and set my job to be retried after the value provided by that header.

cpu:
@jeromegn I'm hopeful we can add it one day. I suspect it will be something we would evaluate as part of the ["v2" implementation](https://letsencrypt.org/2017/06/14/acme-v2-api.html) for Jan 2018. Apologies for being the bearer of bad news!

unixcharles:
I don't think it hurt adding it already. Any opposition?

parkr:
@unixcharles I would love to use this – any interest on merging & cutting a new release?

cpu:
@parkr The `Retry-After` header is still unimplemented in Boulder. Are you targeting a different ACME server than one run by Let's Encrypt?

unixcharles:
@cpu its the reason why it was never merged.

> @unixcharles I would love to use this – any interest on merging & cutting a new release?

I never merged because its not implemented in boulder but if you have a different server implementation that supports it, I don't have any strong opposition.

jeromegn:
Do we know if this has been implemented by boulder when using ACME v2?

cpu:
> Do we know if this has been implemented by boulder when using ACME v2?

Boulder doesn't implement `Retry-After` for V1 or V2 at the present time.

PR #129, opened 2017-12-25
Do not cache error type

eagletmt:
acme-client reports wrong exceptions from the second error response
because Faraday middleware is re-used.

```ruby
client = Acme::Client.new(...)
begin
  client.new_certificate(...)
rescue Acme::Client::Error::Unauthorized
  # When #authorize fails due to some reason, the exception is always
  # set to Acme::Client::Error::Unauthorized wrongly.
  client.authorize(...)
end
```


unixcharles:
Good catch @eagletmt, thanks for contributing.

PR #134, opened 2018-04-10
Acme v2

unixcharles:
Lot of changes here.

Most of the concept remain the same but its almost a complete rewrite for v2. Some API change were made for both consistency but also to stay as close as possible to the specification.

Notable design changes

* Resource objects are now easy to serialize and re-instantiate.
* Resource objects can all be retrieve from URL.
* Resource objects delegate API requests back to the client itself.
* The client does not make any assumption on API endpoints and retrieve everything from the directory URL.
* Group error, client errors vs server errors.

I also re-written documentation / readme.

The API goal for the gem is to stay close to the protocol itself while having this Ruby feeling that make it easy to work with.


README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
+
+## Obtaining a certificate
+### Ordering a certificate
 
-# We need an ACME server to talk to, see github.com/letsencrypt/boulder
-# WARNING: This endpoint is the production endpoint, which is rate limited and will produce valid certificates.
-# You should probably use the staging endpoint for all your experimentation:
-# endpoint = 'https://acme-staging.api.letsencrypt.org/'
-endpoint = 'https://acme-v01.api.letsencrypt.org/'
+To order a new certificate, the client must provider a list of identifiers.
```
will-in-wi: "provider" to "provide"?

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
 
-# Initialize the client
-require 'acme-client'
-client = Acme::Client.new(private_key: private_key, endpoint: endpoint, connection_options: { request: { open_timeout: 5, timeout: 5 } })
+The returned order will contains a list of `Authorization` that need to be completed in other to finalize the order, generally one per identifiers.
```
will-in-wi: "contains" to "contain"?

Since `Authorization` is singular and the context is plural, would it make sense to phrase it like "…contain a list of `Authorization` objects that need…"

"identifiers" to "identifier"?

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
 
-# If the private key is not known to the server, we need to register it for the first time.
-registration = client.register(contact: 'mailto:contact@example.com')
+Each authorization contains multiple challenges, typically a `dns-01` and a `http-01` challenge. The applicant is only required to complete one of the two challenge.
```
will-in-wi: "challenge" to "challenges"

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
+challenge = authorization.http
 ```
 
-### Authorize for domain
+### Preparing for HTTP challenge
+
+To complete the HTTP challenge, you must prepare return a file using HTTP.
 
-Before you are able to obtain certificates for your domain, you have to prove that you are in control of it.
+The path follow the following format:
```
will-in-wi: "follow" to "follows"

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
+```ruby
+order = client.new_order(identifiers: ['example.com'])
+authorization = order.authorizations.first
+challenge = authorization.http
 ```
 
-### Authorize for domain
+### Preparing for HTTP challenge
+
+To complete the HTTP challenge, you must prepare return a file using HTTP.
```
will-in-wi: Should "prepare" be removed? E.g. "…you must return a file using HTTP."

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
 
-Before you are able to obtain certificates for your domain, you have to prove that you are in control of it.
+The path follow the following format:
+
+> .well-known/acme-challenge/#{token}
+
+And the file content is the key authorization. The HTTP01 object as utility methods to generate them.
```
will-in-wi: "The HTTP01 object **has** utility methods"

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
+> http_challenge.content_type # => 'text/plain'
+> http_challenge.file_content # => example_token.TO1xJ0UDgfQ8WY5zT3txynup87UU3PhcDEIcuPyw4QU
+> http_challenge.filename # => '.well-known/acme-challenge/example_token'
+> http_challenge.token # => 'example_token'
+```
 
-# If authorization.status returns 'valid' here you can already get a certificate
-# and _must not_ try to solve another challenge.
-authorization.status # => 'pending'
+For test purpose you can just save the challenge file and use Ruby to serve it:
```
will-in-wi: "purpose" to "purposes"

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
-# This example is using the http-01 challenge type. Other challenges are dns-01 or tls-sni-01.
-challenge = authorization.http01
+### Preparing for DNS challenge
 
-# The http-01 method will require you to respond to a HTTP request.
+To complete the DNS challenge, you must set a DNS record to prove that you control the domain.
 
-# You can retrieve the challenge token
-challenge.token # => "some_token"
+The DNS01 object as utility methods to generate them.
```
will-in-wi: "DNS01 object **has** utility methods"

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
 
-# You can retrieve the expected path for the file.
-challenge.filename # => ".well-known/acme-challenge/:some_token"
+```ruby
+> dns_challenge.record_name # => '_acme-challenge'
+> dns_challenge.record_type # => 'TXT'
+> dns_challenge.record_content # => 'HRV3PS5sRDyV-ous4HJk4z24s5JjmUTjcCaUjFt28-8'
```
will-in-wi: You might not want the `> ` at the start of the lines, since it inhibits copying and is inconsistent with the rest of the README.

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
+```
 
-# You can generate the body of the expected response.
-challenge.file_content # => 'string token and JWK thumbprint'
+### Request a challenge verification
 
-# You are not required to send a Content-Type. This method will return the right Content-Type should you decide to include one.
-challenge.content_type
+Once your are ready to complete the challenge, you can request to server to perform the verification.
```
will-in-wi: "your are" to "you are"

Perhaps "you can request the server perform the verification"

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
+### Downloading a certificate
 
-# Once you are ready to serve the confirmation request you can proceed.
-challenge.request_verification # => true
-challenge.authorization.verify_status # => 'pending'
+Once all required authorizations have been validated through challenges, the order can be finalized using a CSR ([Certificate Signing Request](https://en.wikipedia.org/wiki/Certificate_signing_request)).
 
-# Wait a bit for the server to make the request, or just blink. It should be fast.
-sleep(1)
+CSR can be slightly tricky to generate using OpenSSL from Ruby standard library. The `acme-client` provide a utility class `CertificateRequest` to help with that.
```
will-in-wi: "A CSR"

Perhaps "`acme-client` provides a utility class" (no "The")

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
 
-# Rely on authorization.verify_status more than on challenge.verify_status,
-# if the former is 'valid' you can already issue a certificate and the status of
-# the challenge is not relevant and in fact may never change from pending.
-challenge.authorization.verify_status # => 'valid'
-challenge.error # => nil
+The certificate generation is happening asynchronously. You may need to poll.
```
will-in-wi: Perhaps "Certificate generation happens asynchronously"?

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
-# If authorization.verify_status is 'invalid', you can get at the error
-# message only through the failed challenge.
-authorization.verify_status # => 'invalid'
-authorization.http01.error # => {"type" => "...", "detail" => "..."}
+```ruby
+csr = Acme::Client::CertificateRequest.new(private_key: private_key, subject: { common_name: 'example.com' })
+order.finalize(csr: csr)
+until order.status != 'processing'
+  sleep(1)
+end
```
will-in-wi: Style preference, but I think you could do this as a one-liner:

```ruby
sleep(1) while order.status == 'processing'
```

Also, `while` might be clearer than `until`, given that there is essentially a double negative between `until` and `!=`.

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
+order.certificate # => PEM format certificate
 ```
 
-### Obtain a certificate
+## Extra
+
+### Certificate revokation
 
-Now that your account is authorized for the domain, you should be able to obtain a certificate for it.
+To revoke a certificate you can call `#revoke` with certificate.
```
will-in-wi: "can call `#revoke` with **the** certificate"

README.md
```diff
@@ -21,135 +21,171 @@ gem 'acme-client'
...
-#   :SSLEnable => true,
-#   :SSLPrivateKey => OpenSSL::PKey::RSA.new( File.read('privkey.pem') ),
-#   :SSLCertificate => OpenSSL::X509::Certificate.new( File.read('cert.pem') )); trap('INT') { s.shutdown }; s.start"
+client.revoke(certificate: certificate)
 ```
 
-# Not implemented
+### Certificate renewal
+
+The is no renewal precess, just create a new order.
```
will-in-wi: "precess" to "process"

lib/acme/client.rb
```diff
@@ -22,15 +21,14 @@ class Acme::Client; end
...
-    'new-authz' => '/acme/new-authz',
-    'new-cert' => '/acme/new-cert',
-    'new-reg' => '/acme/new-reg',
-    'revoke-cert' => '/acme/revoke-cert'
-  }.freeze
-
-  def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory_uri: nil, connection_options: {})
+  DEFAULT_DIRECTORY = 'http://127.0.0.1:4000/directory'.freeze
+  repo_url = 'https://github.com/unixcharles/acme-client'
+  USER_AGENT = "Acme::Client v#{Acme::Client::VERSION} (#{repo_url})".freeze
```
will-in-wi: It looks like `repo_url` is only used in `USER_AGENT`. If this is the case, would it make more sense to just interpolate it?

lib/acme/client.rb
```diff
@@ -41,93 +39,226 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
+  def new_account(contact:, terms_of_service_agreed: nil)
     payload = {
-      resource: 'new-reg', contact: Array(contact)
+      contact: Array(contact)
     }
 
-    response = connection.post(@operation_endpoints.fetch('new-reg'), payload)
-    ::Acme::Client::Resources::Registration.new(self, response)
+    if terms_of_service_agreed
+      payload['termsOfServiceAgreed'] = terms_of_service_agreed
```
will-in-wi: Why are the the keys strings here and symbols in `#account_update`?

lib/acme/client.rb
```diff
@@ -41,93 +39,226 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
-  def new_certificate(csr)
-    payload = {
-      resource: 'new-cert',
-      csr: Base64.urlsafe_encode64(csr.to_der)
-    }
+  def account
+    if @kid.nil?
+      response = post(endpoint_for('newAccount'), payload: { onlyReturnExisting: true }, mode: :jwk)
+      @kid = response.headers.fetch(:location)
+    end
```
will-in-wi: What about

```ruby
@kid ||= begin
  response = post(endpoint_for('newAccount'), payload: { onlyReturnExisting: true }, mode: :jwk)
  response.headers.fetch(:location)
end
```

or even

```ruby
@kid ||= begin
  post(endpoint_for('newAccount'), payload: { onlyReturnExisting: true }, mode: :jwk)
    .headers
    .fetch(:location)
end
```

lib/acme/client.rb
```diff
@@ -41,93 +39,226 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
-      configuration.adapter Faraday.default_adapter
+  def order(url:)
+    response = get(url)
+    arguments = Acme::Client::Resources::Order.arguments_from_response(response)
+    Acme::Client::Resources::Order.new(self, **arguments)
+  end
+
+  def finalize(url:, csr:)
+    unless csr.respond_to?(:to_der)
+      raise ArgumentError, 'csr must respond to `to_der`'
```
will-in-wi: Maybe change the error message to `#to_der`?

lib/acme/client.rb
```diff
@@ -41,93 +39,226 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
     end
+
+    base64_der_csr = Acme::Client::Util.urlsafe_base64(csr.to_der)
+
+    payload = { csr: base64_der_csr }
+    response = post(url, payload: payload)
```
will-in-wi: Why not inline payload?

lib/acme/client.rb
```diff
@@ -41,93 +39,226 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
+    arguments = Acme::Client::Resources::Order.arguments_from_response(response)
+    Acme::Client::Resources::Order.new(self, **arguments)
+  end
+
+  def certificate(url:)
+    response = download(url)
+    response.body
```
will-in-wi: This could be a one-liner `download(url).body`

lib/acme/client.rb
```diff
@@ -41,93 +39,226 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
+
+  def authorization(url:)
+    response = get(url)
+    arguments = Acme::Client::Resources::Authorization.arguments_from_response(response)
+    Acme::Client::Resources::Authorization.new(self, url: url, **arguments)
+  end
+
+  def deactivate_authorization(url:)
+    payload = { status: 'deactivated' }
+    response = post(url, payload: payload)
```
will-in-wi: Maybe inline `payload`?

lib/acme/client.rb
```diff
@@ -41,93 +39,226 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
+
+  def challenge(url:)
+    response = get(url)
+    arguments = Acme::Client::Resources::Challenges.arguments_from_response(response)
+    Acme::Client::Resources::Challenges.new(self, **arguments)
+  end
+
+  def request_challenge_validation(url:, key_authorization:)
+    payload = { 'keyAuthorization' => key_authorization }
+    response = post(url, payload: payload)
```
will-in-wi: Maybe inline `payload`?

lib/acme/client/faraday_middleware.rb
```diff
@@ -35,6 +37,12 @@ def on_complete(env)
 
   private
 
+  def jws_header
+    headers = { nonce: pop_nonce, url: env.url.to_s }
+    headers[:kid] = client.kid if @mode == :kid
+    headers
```
will-in-wi: Style preference, but you could write this as:

```ruby
{ nonce: pop_nonce, url: env.url.to_s }.tap do |headers|
  headers[:kid] = client.kid if @mode == :kid
end
```

lib/acme/client/jwk/base.rb
```diff
@@ -56,12 +56,13 @@ def thumbprint
...
+  def jws_header(header)
+    jws = {
       typ: 'JWT',
-      alg: jwa_alg,
-      jwk: to_h
-    }
+      alg: jwa_alg
+    }.merge(header)
+    jws[:jwk] = to_h if header[:kid].nil?
+    jws
```
will-in-wi: Similarly, you could do:
```ruby
{
  typ: 'JWT',
  alg: jwa_alg
}.merge(header).tap do |jws|
  jws[:jwk] = to_h if header[:kid].nil?
end
```

lib/acme/client/resources/authorization.rb
```diff
@@ -1,44 +1,86 @@
...
-                    @http01 ||= HTTP01.new(self)
-                  when 'dns-01'
-                    @dns01 ||= DNS01.new(self)
-                  when 'tls-sni-01'
-                    @tls_sni01 ||= TLSSNI01.new(self)
-      end
-
-      challenge.assign_attributes(attributes) if challenge
-    end
+  def initalize_challenge(attributes)
```
will-in-wi: `initalize_challenge` should probably be `initialize_challenge`

lib/acme/client/resources/authorization.rb
```diff
@@ -1,44 +1,86 @@
...
+  def challenges
+    @challenges.map do |challenge|
+      initalize_challenge(challenge)
+    end
+  end
+
+  def http01
+    @http01 = challenges.find { |challenge|
+      challenge.is_a?(Acme::Client::Resources::Challenges::HTTP01)
+    }
```
will-in-wi: There is a mix of styles using `do; end` vs. `{ }`. Should this and `#dns01` use `do; end`?

lib/acme/client/resources/challenges/base.rb
```diff
@@ -1,39 +1,43 @@
...
-    @authorization = authorization
+  def initialize(client, **arguments)
+    @client = client
+    assign_attributes(arguments)
   end
 
-  def client
-    authorization.client
+  def challenge_type
+    self.class::CHALLENGE_TYPE
```
will-in-wi: Where is `CHALLENGE_TYPE` set? I don't see `Acme::Client::Resources::Challenges::Base::CHALLENGE_TYPE` defined…

spec/spec_helper.rb
```diff
@@ -1,6 +1,11 @@
 $LOAD_PATH.unshift File.join(__dir__, '../lib')
 $LOAD_PATH.unshift File.join(__dir__, 'support')
 
+require 'openssl'
+
+$directory_url = ENV['ACME_DIRECTORY_URL'] || 'https://127.0.0.1/directory'
```
will-in-wi: Could this be a method instead of a global? It totally works but the mere existence of a global feels dirty to me. At least a global method can't be mutated.

spec/support/ssl_helper.rb
```diff
@@ -26,7 +26,8 @@ def generate_key
       when 2
         generate_ecdsa_key('secp384r1')
       when 3
-        generate_ecdsa_key('secp521r1')
+        # generate_ecdsa_key('secp521r1')
```
will-in-wi: Is this comment needed?

lib/acme/client.rb
unixcharles: Its just to keep the line short

lib/acme/client/faraday_middleware.rb
unixcharles: I think its a bit more readable to just return the variable

lib/acme/client/resources/authorization.rb
unixcharles: There is a bug here because I intended to do `||=`.

On this project I've been following the convention of using `{ }` when the return value and `do / end` when the side effect is desired.

lib/acme/client/resources/challenges/base.rb
unixcharles: `Base` is an abstract class ,`CHALLENGE_TYPE` is expected to be defined in the child classes. This is also why I do the `self.class::` dance because otherwise it would attempt to find it in the scope of `Base`.

will-in-wi: Ah, that totally makes sense.

spec/support/ssl_helper.rb
unixcharles: These keys where failing against staging when I started developing. I'll give it another go.

@cpu do you know anything about that?

cpu: @unixcharles Is this something that was working and broke with your V2 work? AFAIK we've never supported secp521r1: https://github.com/letsencrypt/boulder/blob/fa5c9176655d9fa8dfca188de08bd5373aca422f/goodkey/good_key.go#L183-L195

unixcharles:
Thanks @will-in-wi for the feedback. I've fixed / reply most comment.

spec/support/ssl_helper.rb
cpu: @unixcharles I don't believe secp521r1 has ever been supported by Let's Encrypt.

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
+```
 
-# If the private key is not known to the server, we need to register it for the first time.
-registration = client.register(contact: 'mailto:contact@example.com')
+If your account is already registered, you can save some API calls by passing your key ID directly. This will avoid an unnecessary API call to retrieve it from your private key.
 
-# You may need to agree to the terms of service (that's up the to the server to require it or not but boulder does by default)
-registration.agree_terms
+```ruby
+client = Acme::Client.new(private_key: private_key, directory: 'https://acme-v01.api.letsencrypt.org/directory', kid: 'https://example.com/acme/acct/1')
```
weppos: Question for the sake of my understanding: I read the ACME v2 specs, and I wonder... is `kid` simply a custom string identifier?

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
+```ruby
+require 'openssl'
+OpenSSL::PKey::RSA.new(File.read('/path/to/private_key.pem'))
+```
+
+See [RSA](https://ruby.github.io/openssl/OpenSSL/PKey/RSA.html) and [EC](https://ruby.github.io/openssl/OpenSSL/PKey/EC.html) for documentation.
+
+
+```ruby
+client = Acme::Client.new(private_key: private_key, directory: 'https://acme-v01.api.letsencrypt.org/directory')
```
weppos: Given this is targeting v2, the directory seems incorrect.

I would actually suggest to use the staging directory in the README as from my experience there is a risk users will just follow the readme blindly using production, without understanding there are strict limits there (and staging should be used for testing).

```
https://acme-staging-v02.api.letsencrypt.org/directory
```

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
 ```
 
 ## Usage
+* [Setting up a client](#setting-up-a-client)
+* [Account management](#account-management)
+* [Obtaining a certificate](#obtaining-a-certificate)
+  * [Ordering a certificate](#ordering-a-certificate)
+  * [Completing an HTTP challenge](#preparing-for-http-challenge)
+  * [Completing an DNS challenge](#preparing-for-dns-challenge)
+  * [Request a challenge verification](#request-a-challenge-verification)
```
davidcornu: "Requesting" would be more consistent

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
+> http_challenge.content_type # => 'text/plain'
+> http_challenge.file_content # => example_token.TO1xJ0UDgfQ8WY5zT3txynup87UU3PhcDEIcuPyw4QU
+> http_challenge.filename # => '.well-known/acme-challenge/example_token'
+> http_challenge.token # => 'example_token'
+```
 
-# If authorization.status returns 'valid' here you can already get a certificate
-# and _must not_ try to solve another challenge.
-authorization.status # => 'pending'
+For test purposes you can just save the challenge file and use Ruby to serve it:
```
davidcornu: testing purposes

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
+dns_challenge.record_content # => 'HRV3PS5sRDyV-ous4HJk4z24s5JjmUTjcCaUjFt28-8'
+```
 
-# You can generate the body of the expected response.
-challenge.file_content # => 'string token and JWK thumbprint'
+### Request a challenge verification
 
-# You are not required to send a Content-Type. This method will return the right Content-Type should you decide to include one.
-challenge.content_type
+Once you are ready to complete the challenge, you can request the server perform the verification.
```
davidcornu: you can request that the server perform a verification

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
+challenge.request_validation
+```
 
-# We'll write the content of the file
-File.write( File.join( 'public', challenge.filename), challenge.file_content )
+The validation is performed asynchronously and can take some time to be performed by the server.
 
-# Optionally save the authorization URI for use at another time (eg: by a background job processor)
-File.write('authorization_uri', authorization.uri)
+You can poll until its status change.
```
davidcornu: changes

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
+### Downloading a certificate
 
-# Once you are ready to serve the confirmation request you can proceed.
-challenge.request_verification # => true
-challenge.authorization.verify_status # => 'pending'
+Once all required authorizations have been validated through challenges, the order can be finalized using a CSR ([Certificate Signing Request](https://en.wikipedia.org/wiki/Certificate_signing_request)).
 
-# Wait a bit for the server to make the request, or just blink. It should be fast.
-sleep(1)
+A CSR can be slightly tricky to generate using OpenSSL from Ruby standard library. `acme-client` provide a utility class `CertificateRequest` to help with that.
```
davidcornu: - the Ruby standard library
- provides a utility class (`CertificateRequest`)

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
 
-# If authorization.verify_status is 'invalid', you can get at the error
-# message only through the failed challenge.
-authorization.verify_status # => 'invalid'
-authorization.http01.error # => {"type" => "...", "detail" => "..."}
+```ruby
+csr = Acme::Client::CertificateRequest.new(private_key: private_key, subject: { common_name: 'example.com' })
+order.finalize(csr: csr)
+sleep(1) while order.status == 'processing'
+order.certificate # => PEM format certificate
```
davidcornu: PEM-formatted certificate

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
-registration.agree_terms
+```ruby
+client = Acme::Client.new(private_key: private_key, directory: 'https://acme-v01.api.letsencrypt.org/directory', kid: 'https://example.com/acme/acct/1')
 ```
 
-### Authorize for domain
+## Account management
 
-Before you are able to obtain certificates for your domain, you have to prove that you are in control of it.
+Account are tied to a private key. Before being allowed to create orders, the account must be registered and the ToS accepted using the private key. The account will be assigned a key ID.
```
etiennebarrie: *Accounts

README.md
```diff
@@ -21,135 +21,169 @@ gem 'acme-client'
...
-# any associated challenges via Acme::Client#fetch_authorization.
-authorization.uri # => '...'
+To order a new certificate, the client must provide a list of identifiers.
 
-# This example is using the http-01 challenge type. Other challenges are dns-01 or tls-sni-01.
-challenge = authorization.http01
+The returned order will contain a list of `Authorization` that need to be completed in other to finalize the order, generally one per identifier.
 
-# The http-01 method will require you to respond to a HTTP request.
+Each authorization contains multiple challenges, typically a `dns-01` and a `http-01` challenge. The applicant is only required to complete one of the two challenges.
```
etiennebarrie: I think we could leave `two` out here.

README.md
weppos: What about an existing account? Can I fetch it in some ways?

lib/acme/client.rb
```diff
@@ -41,93 +39,221 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
+    body = fetch_directory
+    @meta = body['meta']
+
+    @directory = {}
+    @directory['newNonce'] = URI(body['newNonce']) if body['newNonce']
+    @directory['newAccount'] = URI(body['newAccount']) if body['newAccount']
+    @directory['newOrder'] = URI(body['newOrder']) if body['newOrder']
+    @directory['newAuthz'] = URI(body['newAuthz']) if body['newAuthz']
+    @directory['revokeCert'] = URI(body['revokeCert']) if body['revokeCert']
+    @directory['keyChange'] = URI(body['keyChange']) if body['keyChange']
```
etiennebarrie: Maybe a loop here?

lib/acme/client/error.rb
```diff
@@ -1,16 +1,32 @@
...
-  class Unauthorized < Acme::Client::Error; end
-  class UnknownHost < Acme::Client::Error; end
   class Timeout < Acme::Client::Error; end
-  class RateLimited < Acme::Client::Error; end
-  class RejectedIdentifier < Acme::Client::Error; end
-  class UnsupportedIdentifier < Acme::Client::Error; end
+
+  class ClientError < Acme::Client::Error; end
+  class InvalidDirectory < ClientError; end
+  class UnsupporedOperation < ClientError; end
```
etiennebarrie: typo: `UnsupportedOperation`, also used in `endpoint_for`

README.md
cpu: @weppos From a protocol standpoint there is a way. You can use the mechanism described in [Section 7.3.1](https://tools.ietf.org/html/draft-ietf-acme-acme-10#section-7.3.1) and POST the `newAccount` endpoint sending a signed message body `{"onlyReturnExisting": true}`. The JWS that authenticates the request should embed the JWK of the account key you want to learn the key ID for. If the key is known on the server side you'll get back the KeyID.

From a `unixcharles/acme-client` perspective I don't know, sorry :-) @unixcharles probably does!

lib/acme/client/resources/account.rb
```diff
@@ -0,0 +1,58 @@
...
+class Acme::Client::Resources::Account
+  attr_reader :url, :status, :contact, :term_of_service, :orders_url
+
+  def self.arguments_from_response(response)
+    attributes = response.body
+    status = attributes.fetch('status')
+    term_of_service = attributes['termsOfServiceAgreed']
+    contact = attributes.fetch('contact', [])
+
+    { term_of_service: term_of_service, status: status, contact: contact }
```
davidcornu: Any reason this doesn't return an instance of `Acme::Client::Resources::Account`?

Seems odd to me to have callers get arguments from a class method to then initialize it separately.

lib/acme/client.rb
```diff
@@ -41,93 +39,221 @@ def initialize(jwk: nil, private_key: nil, endpoint: DEFAULT_ENDPOINT, directory
...
-    response = connection.post(@operation_endpoints.fetch('new-reg'), payload)
-    ::Acme::Client::Resources::Registration.new(self, response)
+    if terms_of_service_agreed
+      payload[:termsOfServiceAgreed] = terms_of_service_agreed
+    end
+
+    response = post(endpoint_for('newAccount'), payload: payload, mode: :jws)
+    @kid = response.headers.fetch(:location)
+
+    if response.body.nil? || response.body.empty?
```
davidcornu: https://tools.ietf.org/html/draft-ietf-acme-acme-11#section-7.3.1

If an account is created, the server responds with a `201`, if not a `200`.

lib/acme/client.rb
davidcornu: You could also move this to a `Directory` class that has methods that return URIs for each of these operations. It could also encapsulate the logic in `endpoint_for` and `fetch_directory`

lib/acme/client/error.rb
```diff
@@ -1,16 +1,32 @@
...
+  class RejectedIdentifier < ServerError; end
+  class ServerInternal < ServerError; end
+  class Unauthorized < ServerError; end
+  class UnsupportedIdentifier < ServerError; end
+  class UserActionRequired < ServerError; end
+  class BadRevocationReason < ServerError; end
+  class Caa < ServerError; end
+  class Dns < ServerError; end
+  class Connection < ServerError; end
+  class Tls < ServerError; end
```
davidcornu: nit - Some of these (`Malformed`, `Caa`, `Dns`, `Tls`) are very ambiguous

lib/acme/client/faraday_middleware.rb
```diff
@@ -67,14 +75,14 @@ def error_name
   end
 
   def error_type_to_klass(type)
-    type.gsub('urn:acme:error:', '').split(/[_-]/).map { |type_part| type_part[0].upcase + type_part[1..-1] }.join
+    type.gsub('urn:ietf:params:acme:error:', '').split(/[_-]/).map { |type_part| type_part[0].upcase + type_part[1..-1] }.join
```
davidcornu: 1. I feel like this logic should `live in lib/acme/client/error.rb`
2. A simple `case` on whatever comes after `urn:ietf:params:acme:error:` would be much simpler to reason about (and would let us have less ambiguous error class names https://github.com/unixcharles/acme-client/pull/134/files#r180874585)
3. If you're against (2), 
    - `gsub` &rarr; `sub` would make it clear there's only one substitution going on
    - `sub(/\Aurn:ietf:params:acme:error:/)` is a bit more intention-revealing seen as `urn:ietf:params:acme:error:` should always be at the beginning

README.md
unixcharles: > From a unixcharles/acme-client perspective I don't know, sorry :-) @unixcharles probably does!

If the `kid` is not provided at initialization, the client will lazy load the `kid` by doing a `POST` to `newAccount` whenever the `kid` is required.

So you can just instantiate  a new client with the private key without the `kid` and call `Acme::Client#kid` on it to retrieve it.

Good input, I'll add instruction for this.

lib/acme/client/resources/account.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def update(contact: nil, terms_of_service_agreed: nil)
+    assign_attributes **@client.account_update(
+      contact: contact, terms_of_service_agreed: term_of_service
+    ).to_h
+    true
+  end
+
+  def deactivate
+    assign_attributes **@client.account_deactivate.to_h
```
davidcornu: If I'm reading this correctly,

1. `@client.account_deactivate` calls `arguments_from_response`
2. The returned hash is used to initialize a `Acme::Client::Resources::Account`
3. We then call `to_h` on that instance to merge those attributes into this instance

This sounds to me like the responsibilities of the `Account` class are spread all over the place and `@client` is doing too much.

lib/acme/client/resources/authorization.rb
```diff
@@ -1,44 +1,86 @@
...
+    assign_attributes **@client.deactivate_authorization(url: url).to_h
+    true
+  end
+
+  def reload
+    assign_attributes **@client.authorization(url: url).to_h
+    true
+  end
+
+  def challenges
```
davidcornu: Given the logic in `http01` and `dns01` I'd recommend introducing a `Challenges` object so that we can do

```ruby
def assign_attributes(...)
  @challenges = Challenges.new(challenges)
end
```

which would let you do

```ruby
authorization.challenges.http01
```

lib/acme/client.rb
unixcharles: I know its a bit weird but [pebble](https://github.com/letsencrypt/pebble) and [boulder](https://github.com/letsencrypt/boulder) are no implemented exactly the same. If I remember correctly pebble return empty body on `201` and expect you to follow the location.

I found that the presence of the location was consistent whenever the body was empty, so that's why I did this. Ideally, the body should always be present.

unixcharles: ^ @cpu I'm getting this right?

lib/acme/client.rb
unixcharles: Move it in an object. I was not sure but its pretty nice actually.

lib/acme/client/error.rb
unixcharles: I map directly to the server error here. I'm not too worried because it come with a clear error message.

lib/acme/client/resources/account.rb
unixcharles: I want each object to be easy serialize/deserialize.

i.e.; I always want to be able to do `ResourceObject.new(client, ResourceObject.to_h)`

Now that you bring it up, I feel like extracting the argument from the response should live in the client instead of the resource objects.

lib/acme/client/resources/account.rb
unixcharles: I'm trying to find a way to allow:

```ruby
client.account_deactivate(url: account_url)
```

But also allow:

```ruby
account = client.account(url: account_url)
account.deactivate
```

But without repeating myself too much. I'm not too worried about the `Client` class getting big, as long as it only do http stuff.

I think that exacting the argument from the response `arguments_from_response` is the flaw in this design, that's what make `account > client > account > client > account`. Do that make sense to you?

lib/acme/client/resources/authorization.rb
unixcharles: Its a nice looking API but I'm not sure if it would bring value. In most scenario you will want to access a specific challenge. i.e.; `authorization.http01`

lib/acme/client/faraday_middleware.rb
unixcharles: Good point. When I came up with this there was no error list in the spec. So I just `#constantize` the error message and kinda defined errors has they appear. I'll replace this with a simple hash in the Error module.

lib/acme/client.rb
cpu: @unixcharles This is in the context of a normal newAccount request correct, not one with `onlyReturnExisting: true`? I don't think this is right if so.

Pebble returns HTTP status `201` **and** the account object for successful `newAccount` requests that create a new account: https://github.com/letsencrypt/pebble/blob/cae1c3522ee7a4871550f4ff9378677e06a4cdf7/wfe/wfe.go#L739

Boulder does the same thing: HTTP status 201 and the account object in the response body: https://github.com/letsencrypt/boulder/blob/3ecf841a3a48b7c504f26e5b040fbfb36c1664b7/wfe2/wfe.go#L531

unixcharles: I think that it happened when attempting to call `newAccount` on an account that was already registered without `onlyReturnExisting: true`.

I'll try to reproduce, maybe I was just doing something wrong with my test setup.

cpu: > I think that it happened when attempting to call newAccount on an account that was already registered without onlyReturnExisting: true.

That would make sense :+1: - There should not be an HTTP Status Created response in this case because no account was created. That also matches what Section 7.3.1 says when you're not using `onlyReturnExisting`:

     If the server already has an account registered with the provided
     account key, then it MUST return a response with a 200 (OK) status
     code and provide the URL of that account in the Location header
     field.  This allows a client that has an account key but not the
     corresponding account URL to recover the account URL.

In that specific case Boulder returns no body and HTTP status OK: https://github.com/letsencrypt/pebble/blob/cae1c3522ee7a4871550f4ff9378677e06a4cdf7/wfe/wfe.go#L686-L693 Pebble does the same: https://github.com/letsencrypt/pebble/blob/cae1c3522ee7a4871550f4ff9378677e06a4cdf7/wfe/wfe.go#L686-L693

unixcharles: Okay, so it would make sense to follow the location headers if not body is returned.

unixcharles:
Thanks for trying it out @nhinds 

I fixed the `Order#reload` bug.

Added `Authorization#identifier` and `Authorization#domain`. This use case just didn't cross my mind since I always do one domain per order/certificate but its pretty legit. The identifier is part of the response but I just didn't assign it anywhere.

@cpu do you happen to know why `"identifier": { "type": "dns", "value": "example.com" }` and not just `"identifier": "example.com"`? Do you know of a case where another type would be possible?

I just assumed its always going to be a `dns` and added a `authorization#domain` attribute to avoid having to dig into the identifier. 

> Also - is the intended way to access the server's terms of service (before agreeing to them) to access client.directory.meta['termsOfService']? Previously the terms of service URI was available on registration.term_of_service_uri.

You are right. Its a bit weird to have to access the directory directly. I added methods for the known meta entries such a `Client#terms_of_service`.

cpu:
> @cpu do you happen to know why "identifier": { "type": "dns", "value": "example.com" } and not just "identifier": "example.com"? 

@unixcharles Protocol extensibility. Boulder/Pebble only implement "dns" type identifiers presently but that may change one day.

> Do you know of a case where another type would be possible?

Follow-up protocol extensions (e.g. `draft-ietf-acme-ip-01`) may define additional identifier types. draft-ietf-acme-ip-01 defines [`"type": "ip"` identifiers](https://tools.ietf.org/html/draft-ietf-acme-ip-01#section-3) as one concrete example.

lib/acme/client/resources/account.rb
unixcharles: I've refactor this [here](https://github.com/unixcharles/acme-client/pull/134/files#diff-e843384610a1a0cb8c21582bbed52188R205)

nhinds:
Thanks - order reloading works and the `.domain` accessor on authorizations is useful.

`order.certificate` seems to function a bit strangely to me. Previously there was a `Acme::Client::Certificate` class which provided separate access to the certificate and its chain. `order.certificate` currently only returns a string containing the PEM-encoded and its chain together, although it's documented in the README as just a "PEM-formatted certificate".

The certificate is downloaded by sending an `Accept: application/pkix-cert` header. The Acme v2 spec indicates that clients can use this header to request a DER-encoded single certificate, although the current servers ignore this and return a response content type `application/pem-certificate-chain` with the full chain (which is valid, as alternative representations are optional and the full chain is the default return value). If any server does eventually implement the `application/pkix-cert` content type, `order.certificate` will suddenly start returning the DER-encoded rather than the PEM-encoded certificate chain.

If the intention is to return the full chain, could `acme-client` pass `Accept: application/pem-certificate-chain` to indicate that's what it is expecting, so that when servers implement `application/pkix-cert` the method does not break?

It would also be nice to continue to provide separate access to the certificate and its chain, although users of the library can probably migrate to doing so themselves e.g. `certificate, chain = order.certificate.split(/(?<=-----END CERTIFICATE-----)/, 2).map(&:strip)`

unixcharles:
> The certificate is downloaded by sending an Accept: application/pkix-cert header. 

Good call, I'll fix the header.

I think that in most scenario, a pem chain is probably what most user need, and the spec make it the default format Its also easy to parse and convert into other format if needed, so I prefer to just return the PEM directly and avoid putting more responsibility on the client.

Does anybody think there would be a good case for wrapping the response?

unixcharles:
Thanks everyone for the review and feedback

PR #152, opened 2018-11-28
Fix splats interpreted as argument prefix warnings

programingnotes:
This PR removes warnings introduced by presence of white space between methods and arguments in:

Acme::Client::Resources::Challenges::Base
Acme::Client::Resources::Authorization
Acme::Client::Resources::Account
Acme::Client::Resources::Order


programingnotes:
@unixcharles : could you please review the pull request? Thank you.

.ruby-version
```diff
@@ -0,0 +1 @@
+2.4.0
```
unixcharles: I would rather not include this change, its specific to the developer environment.

lib/acme/client/resources/order.rb
```diff
@@ -8,8 +8,12 @@ def initialize(client, **arguments)
     assign_attributes(arguments)
   end
 
+  def get_order
+    @client.order(url: url).to_h
+  end
```
unixcharles: Same with this method with this method.

spec/jwk_spec.rb
```diff
@@ -75,7 +75,7 @@
       it 'is not ASN.1 encoded' do
         expect {
           OpenSSL::ASN1.decode(subject.sign(message))
-        }.to raise_error
+        }.to raise_error(OpenSSL::ASN1::ASN1Error)
```
unixcharles: It's a bit of an unrelated change but being specific on the expected error does not hurt.

unixcharles:
Thanks for the PR. I've never ran this project with `-w` to be honest, so I didn't notice.

I'm OK with adding the parenthesis where used with spat. It remove potential confusion of not knowing if you are multiplying or passing spat arguments.

PR #153, opened 2019-02-07
Fix Acme::Client connection to use with proxy

hiphref:
Hi! When I use `Acme::Client.new(private_key:pk, directory: dir, connection_options: {proxy: 'proxy'})` i get error because get_nonce method create new `Faraday.head` connection without `connection_options`.

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -178,7 +178,7 @@ class Acme::Client
 
   def get_nonce
     connection = new_connection(endpoint: endpoint_for(:new_nonce))
-    response = connection.head(nil, nil, 'User-Agent' => Acme::Client::USER_AGENT)
+    response = connection.head(nil, nil, 'User-Agent' => USER_AGENT)
     nonces << response.headers['replay-nonce']
     true
   end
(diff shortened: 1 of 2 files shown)


lib/acme/client.rb
```diff
@@ -268,12 +269,18 @@ def connection_for(url:, mode:)
...
-    Faraday.new(endpoint, **@connection_options) do |configuration|
+  def new_acme_connection(endpoint:, mode:)
+    new_connection(endpoint: endpoint) do |configuration|
       configuration.use Acme::Client::FaradayMiddleware, client: self, mode: mode
+    end
+  end
+
+  def new_connection(endpoint:, &block)
+    Faraday.new(endpoint, **@connection_options) do |configuration|
+      yield(configuration) if block
```
jhass: No need to capture the block, just `if block_given?`

lib/acme/client.rb
```diff
@@ -177,7 +177,8 @@ def revoke(certificate:, reason: nil)
   end
 
   def get_nonce
-    response = Faraday.head(endpoint_for(:new_nonce), nil, 'User-Agent' => USER_AGENT)
+    connection = new_connection(endpoint: endpoint_for(:new_nonce))
+    response = connection.head(nil, nil, 'User-Agent' => Acme::Client::USER_AGENT)
```
jhass: Why did you switch to the fully qualified constant?

lib/acme/client.rb
hiphref: 👍

PR #158, opened 2019-03-22
Update README.md and mention using jwk's (closes #156

### LINKED ISSUE #156: Mention Client.new's jwk: keyword
I've actually spent a few days trying to find a way to convert to pem just to find out by browsing the source that you already implemented that way. It would be nice if the jwk keyword was mentioned in the readme.

IotaSpencer:
This mentions using jwk's as private keys instead of just a 'pem' file

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,24 +52,46 @@ require 'openssl'
 private_key = OpenSSL::PKey::RSA.new(4096)
 ```
 
-Or load one from a PEM file
+Load one from a PEM file
 
 ```ruby
 require 'openssl'
 OpenSSL::PKey::RSA.new(File.read('/path/to/private_key.pem'))
 ```
-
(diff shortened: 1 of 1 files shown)


unixcharles:
I don't know about this.

This project does not use the `jose` gem. The fact that this works is a coincidence where `JOSE::JWK` and `Acme::Client::JWK::Base` happen to be implemented with the same interface.

The reason why there is an argument to pass a JWK object, is to allow the users of the library to implement their own JWK that conform to [jwk/base](https://github.com/unixcharles/acme-client/blob/master/lib/acme/client/jwk/base.rb).

For full context you can look into the initial implementation: https://github.com/unixcharles/acme-client/pull/120

jhass:
OTOH the `jose` gem seems popular enough, we could consider shipping an adapter, `Acme::Client::JWK::JOSE.new(a_jose_jwk)` that confirms to our interface (and just delegates for now), then document usage of that.

unixcharles:
Cleaning up. Will stick to the idea of no adding deps on external gems other than faraday.

Issue #148, opened 2018-08-09
Nonces expiration and Acme::Client::Error::BadNonce


cema-sp:
Greetings!
We are using **acme-client** to update certificated of managed domains in an automated way.
We have created a daemon that sits in memory and recurrently orders certificates. It memoizes `client` instance and re-uses it on every scheduled run.

The `client` [stores anti-reply nonces in memory](https://github.com/unixcharles/acme-client/blob/38f4e17bc9e4ec21c3a417a085d843d1fb7f8b4d/lib/acme/client.rb#L44) and uses then on demand.

What we have noticed is that Letsencrypt nonces could (most probably) expire after approximately 24 hours, which results into the following error

~~~
Acme::Client::Error::BadNonce: JWS has an invalid anti-replay nonce
~~~

Because of that reason we had to refrain from memoizing the `client`.

Do you think it would be useful to allow **acme-client** users to provide something like a "nonce store instance"? It could be optional and configurable, and would allow one to store nonces keeping in mind their expiration.

~~~ruby
# lib/acme/client.rb
...
  def initialize(jwk: nil, kid: nil, private_key: nil, directory: DEFAULT_DIRECTORY, connection_options: {}, nonce_store: [])
    ...
    @nonces ||= nonce_store
  end
...

# my_client.rb

my_store = StoreWithExpiration.new
client = Acme::Client.new(nonce_store: my_store)
~~~

unixcharles:
What about just retrying on `Acme::Client::Error::BadNonce`?

Nonce expiration is not really part of the spec, so I prefer to make no assumption about it and consider it an implementation details. I would generally consider `BadNonce` as retry-able.

cema-sp:
@unixcharles Thank you for answering, I'll try this approach.

cpu:
> Nonce expiration is not really part of the spec, so I prefer to make no assumption about it and consider it an implementation details. I would generally consider BadNonce as retry-able.

Indeed, and retrying on badNonce **is** [in the spec](https://tools.ietf.org/html/draft-ietf-acme-acme-13#section-6.4)!
> An error response with the "badNonce" error type MUST include a Replay-Nonce header with a fresh    nonce.  On receiving such a response, a client SHOULD retry the request using the new nonce.

:+1: 

> What we have noticed is that Letsencrypt nonces could (most probably) expire after approximately 24 hours, which results into the following error

[...]

PR #163, opened 2019-08-07
Faraday: Retry BadNonce errors at the connection level.

zmack:
Given that the spec states that clients should be retrying these ( https://tools.ietf.org/html/draft-ietf-acme-acme-13#section-6.4 ) it makes sense to deal with retries on the client level rather than
relying on the consumers to deal with it.

I know that there was a discussion around this here https://github.com/unixcharles/acme-client/issues/148, just seems that it's easier to have this be transparent to clients.

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -29,7 +29,7 @@ class Acme::Client
     pem: 'application/pem-certificate-chain'
   }
 
-  def initialize(jwk: nil, kid: nil, private_key: nil, directory: DEFAULT_DIRECTORY, connection_options: {})
+  def initialize(jwk: nil, kid: nil, private_key: nil, directory: DEFAULT_DIRECTORY, connection_options: {}, bad_nonce_retry: 0)
     if jwk.nil? && private_key.nil?
       raise ArgumentError, 'must specify jwk or private_key'
     end
@@ -41,6 +41,7 @@ class Acme::Client
     end
 
     @kid, @connection_options = kid, connection_options
+    @bad_nonce_retry = bad_nonce_retry
     @directory = Acme::Client::Resources::Directory.new(URI(directory), @connection_options)
     @nonces ||= []
   end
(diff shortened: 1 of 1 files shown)


zmack:
Oh welp, was expecting to get some other change to have a bit of a preface to this PR.

First up, thanks for all your work on this client, it's been great in all of our work with Let's Encrypt.

A recent [incident](https://letsencrypt.status.io/pages/incident/55957a99e800baa4470002da/5d40614b3df4b70c69ed16dd) on let's encrypt's side made our integration break and we ended up monkeypatching the client to get transparent retries in.

As far as I could tell the main reason ended up being the fact that unless persistent connections are used, you run the risk of ping-pong-ing between nodes where one node won't recognize the nonce from the other. Automatically retrying alleviates this, and it requires no changes from the library's clients.

unixcharles:
I originally decided let those bubble up because in general bad BadNonce are pretty rare and it's likely that something fishy is happening on the clients. I planned to use the library in a managed background queue like sidekiq that already handle retry on exceptions.

It would be a bit of a departure from the current design which basically bubble up all server errors.

I understand your use case and it's pretty legit. I suppose you could pass all those params to `connection_options` when initializing the client to avoid monkey patching and they would get forwarded to the client initialization. It's just not really user friendly for a common use case like this.

What about adding something like `bad_nonce_retry: 3` as an option kwarg for the client?

zmack:
Updated the PR, was that what you had in mind ?

zmack:
I had omitted to pay the rubocop tax. Sorted that all out now 😀

zmack:
Just checking in on this. @unixcharles, is there anything you'd like me to change here?

Thank you.

unixcharles:
If you are comfortable adding test and record a VCR fixture for it, that would be nice.

Otherwise, I will write a test for it  before merging, probably over the weekend.

unixcharles:
Feature released in 2.0.4. Thanks for the PR @zmack!

PR #165, opened 2019-08-21
Retry non bad nonce

unixcharles:
tweak https://github.com/unixcharles/acme-client/pull/163

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -29,7 +29,7 @@ class Acme::Client
     pem: 'application/pem-certificate-chain'
   }
 
-  def initialize(jwk: nil, kid: nil, private_key: nil, directory: DEFAULT_DIRECTORY, connection_options: {})
+  def initialize(jwk: nil, kid: nil, private_key: nil, directory: DEFAULT_DIRECTORY, connection_options: {}, bad_nonce_retry: 0)
     if jwk.nil? && private_key.nil?
       raise ArgumentError, 'must specify jwk or private_key'
     end
@@ -41,6 +41,7 @@ class Acme::Client
     end
 
     @kid, @connection_options = kid, connection_options
+    @bad_nonce_retry = bad_nonce_retry
     @directory = Acme::Client::Resources::Directory.new(URI(directory), @connection_options)
     @nonces ||= []
   end
@@ -280,12 +281,12 @@ class Acme::Client
 
   def new_connection(endpoint:)
     Faraday.new(endpoint, **@connection_options) do |configuration|
-      configuration.request(:retry, max: 10,
+      configuration.request(:retry, max: @bad_nonce_retry,
                             interval: 0.05,
                             interval_randomness: 0.5,
                             backoff_factor: 2,
                             methods: Faraday::Connection::METHODS,
-                            exceptions: [Acme::Client::Error::BadNonce])
+                            exceptions: [Acme::Client::Error::BadNonce]) if @bad_nonce_retry > 0
       yield(configuration) if block_given?
       configuration.adapter Faraday.default_adapter
     end
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -282,13 +282,13 @@ class Acme::Client
   def new_connection(endpoint:)
     Faraday.new(endpoint, **@connection_options) do |configuration|
       if @bad_nonce_retry > 0
-      configuration.request(:retry,
-        max: @bad_nonce_retry,
-        interval: 0.05,
-        interval_randomness: 0.5,
-        backoff_factor: 2,
-        methods: Faraday::Connection::METHODS,
-        exceptions: [Acme::Client::Error::BadNonce])
+        configuration.request(:retry,
+          max: @bad_nonce_retry,
+          interval: 0.05,
+          interval_randomness: 0.5,
+          backoff_factor: 2,
+          methods: Faraday::Connection::METHODS,
+          exceptions: [Acme::Client::Error::BadNonce])
       end
       yield(configuration) if block_given?
       configuration.adapter Faraday.default_adapter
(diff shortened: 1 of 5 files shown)


unixcharles:
@zmack, does that work for you? I removed the options that were set to default and wrote a test case.

zmack:
Thank you so much for adding the tests!

Regarding the options, those seem to default to:
```ruby
interval: 0,
interval_randomness: 0,
backoff_factor: 1,
```
where the retry interval function is
```ruby
def calculate_retry_interval(retries)
  retry_index = @options.max - retries
  current_interval = @options.interval * (@options.backoff_factor ** retry_index)
  current_interval = [current_interval, @options.max_interval].min
  random_interval  = rand * @options.interval_randomness.to_f * @options.interval

  current_interval + random_interval
end
```
which looks like it means that it'll always just straight up retry right away. 

Not having any sort of backoff might be a bit problematic, but I can't say i feel very strongly about this, given that it shouldn't be too frequent of an occurrence

unixcharles:
I would instinctively assume that `badNonce` should be retried without waiting.

I'm a bit worried about adding any kind of sleep without the user setting it explicitly. It's too easy to have one part of your system fail and start consuming all the resource.

zmack:
> I'm a bit worried about adding any kind of sleep without the user setting it explicitly.

Yeah, that's a fair point.

unixcharles:
Released in 2.0.4

weppos:
@unixcharles @zmack I have a question regarding this change.

By default, the Retry middleware is defined to retry Timeout errors.
https://github.com/lostisland/faraday/blob/master/lib/faraday/request/retry.rb#L5-L24

By setting an explicit class of exceptions to retry, you are overriding the default settings, effectively invalidating the ability of the client to retry Timeouts.

```
    # Catches exceptions and retries each request a limited number of times.
    #
    # By default, it retries 2 times and handles only timeout exceptions. It can
    # be configured with an arbitrary number of retries, a list of exceptions to
    # handle, a retry interval, a percentage of randomness to add to the retry
    # interval, and a backoff factor.
    #
    # @example Configure Retry middleware using intervals
    #   Faraday.new do |conn|
    #     conn.request(:retry, max: 2,
    #                          interval: 0.05,
    #                          interval_randomness: 0.5,
    #                          backoff_factor: 2,
    #                          exceptions: [CustomException, 'Timeout::Error'])
    #
    #     conn.adapter(:net_http) # NB: Last middleware must be the adapter
    #   end
    #
    # This example will result in a first interval that is random between 0.05
    # and 0.075 and a second interval that is random between 0.1 and 0.15.
```

In other words, when `bad_nonce_retry` in the ACME client is set, the client won't retry Timeouts anymore. Vice-versa, by default it will.

So my question is: was this intended, is it a design decision, or it's a not planned side effect and it may make sense to have the `bad_nonce_retry` setting to augment the default retry capabilities, rather than replacing them?

unixcharles:
I would say its by design. The option is `bad_nonce_retry`, I see the fact the it use the retry middleware is an implementation detail.

I wouldn't be against the idea of having a `timeout_retry` option but as a user I would find `bad_nonce_retry` having the side effect of retrying timeout errors very unexpected.

In my opinion, timeout recovery should be something that is handled by the application.

If latency was to go up on the servers at LetsEncrypt, its relatively easy for the application to start exhausting its resources, so for this reason I put low timers. If there was a retry, my timeout would actually be multiplied by the number retries unexpectedly.

weppos:
> In my opinion, timeout recovery should be something that is handled by the application.

@unixcharles then shouldn't it always be disabled? Because now if you don't pass `bad_nonce_retry` it will actually enable the Timeout retries by default.

unixcharles:
If you don't pass anything, `bad_nonce_retry` will default to `0` and the middleware will not be injected

```
def new_connection(endpoint:)
  Faraday.new(endpoint, **@connection_options) do |configuration|
    if @bad_nonce_retry > 0
      configuration.request(:retry,
        max: @bad_nonce_retry,
        methods: Faraday::Connection::METHODS,
        exceptions: [Acme::Client::Error::BadNonce])
    end
    yield(configuration) if block_given?
    configuration.adapter Faraday.default_adapter
  end
end
```

I'm missing something?

weppos:
I was under the impression the middleware was automatically injected on initialization, but after a deeper verification I believe it's not:

```
irb(main):007:0> f = Faraday.new { |x| x.request(:retry, max: 2) }
=> #<Faraday::Connection:0x00007f97a84567c0 @parallel_manager=nil, @headers={"User-Agent"=>"Faraday v0.15.4"}, @params={}, @options=#<Faraday::RequestOptions (empty)>, @ssl=#<Faraday::SSLOptions (empty)>, @default_parallel_manager=nil, @builder=#<Faraday::RackBuilder:0x00007f97a84558c0 @handlers=[Faraday::Request::Retry]>, @url_prefix=#<URI::HTTP http:/>, @manual_proxy=false, @proxy=nil, @temp_proxy=nil>
irb(main):008:0> f = Faraday.new
=> #<Faraday::Connection:0x00007f97a4c9db30 @parallel_manager=nil, @headers={"User-Agent"=>"Faraday v0.15.4"}, @params={}, @options=#<Faraday::RequestOptions (empty)>, @ssl=#<Faraday::SSLOptions (empty)>, @default_parallel_manager=nil, @builder=#<Faraday::RackBuilder:0x00007f97a4c9cd70 @handlers=[Faraday::Request::UrlEncoded, Faraday::Adapter::NetHttp]>, @url_prefix=#<URI::HTTP http:/>, @manual_proxy=false, @proxy=nil, @temp_proxy=nil>
```

You can disregard my last comment then. Sorry for that.

unixcharles:
No worry, thanks looking into it.

PR #169, opened 2019-09-25
Add an ugly fix for a problem with the api

### LINKED ISSUE #168: Faraday leaves ports away in JWS headers
We are hitting a strange problem with the Let's Encrypt API at the moment.

We're always getting these errors for several of the urls. It's really weird, because the API clearly returns the URL's without the port. Anyway we will have to add the port to the url when signing the request in the Faraday middleware.

```
Error: JWS header parameter 'url' incorrect. Expected "https://acme-v02.api.letsencrypt.org:443/acme/new-order" got "https://acme-v02.api.letsencrypt.org/acme/new-order"
```

We are trying to find a way around that and will come back with a PR if we find something. 

URI::Generic#to_s leaves the port away if it's 443. That shouldn't happen.

https://github.com/unixcharles/acme-client/blob/master/lib/acme/client/faraday_middleware.rb#L44

neckhair:
acme-client receives the endpoints to create orders and similar stuff from the let's encrypt (boulder) `/directory` endpoint. Those urls don't include the port (443).

When querying the boulder API it expects some signed headers (JWS). In those headers Boulder expects these urls to include the port. At least that's what we recently got:

```
JWS header parameter 'url' incorrect. Expected "https://acme-v02.api.letsencrypt.org:443/acme/new-order" got "https://acme-v02.api.letsencrypt.org/acme/new-order"
```

Ruby's URI::Generic#to_s, which is used in the Faraday middleware omits the port if it's the standard port for the url scheme and there is no way to add it back other than doing some ugly string concatenations.

I'm aware that this is an ugly solution. It helped us to restore our production systems. But maybe someone has a better way to fix this or knows what changed on Let's Encrypt's side and if there is a better solution than my code?

Fixes #168

### DIFF
--- a/lib/acme/client/faraday_middleware.rb
+++ b/lib/acme/client/faraday_middleware.rb
@@ -38,7 +38,9 @@ class Acme::Client::FaradayMiddleware < Faraday::Middleware
   private
 
   def jws_header
-    headers = { nonce: pop_nonce, url: env.url.to_s }
+    # hack to include the port which boulder expects but URI::HTTPS#to_s omits
+    url = "#{env.url.scheme}://#{env.url.hostname}:#{env.url.port}#{env.url.path}"
+    headers = { nonce: pop_nonce, url: url }
     headers[:kid] = client.kid if @mode == :kid
     headers
   end
--- a/lib/acme/client/faraday_middleware.rb
+++ b/lib/acme/client/faraday_middleware.rb
@@ -38,7 +38,7 @@ class Acme::Client::FaradayMiddleware < Faraday::Middleware
   private
 
   def jws_header
-    # hack to include the port which boulder expects but URI::HTTPS#to_s omits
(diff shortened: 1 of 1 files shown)


neckhair:
I also opened an issue at boulder to get some feedback about what has changed: https://github.com/letsencrypt/boulder/issues/4447

unixcharles:
Hey @cpu, it doesn't seem like we have any indication for what would be the appropriate behaviour in the spec?

Should we use the url given by the directory as is, or remove default ports like 443 for https (or 80 for http lol)?

neckhair:
Looks like this PR shouldn't be merged. See https://github.com/letsencrypt/boulder/issues/4447. But the branch can be used to get a temporary workaround. I keep this open until it's solved on Let's Encrypt's side. So that people find it more easily.

PR #170, opened 2019-10-10
Use post-as-get for account, authz, certificate, challenge, and order…

github1:
This change addresses API changes in pebble 2.0.0 (described in https://github.com/letsencrypt/pebble/issues/192) primarily around requiring POST-as-GET instead of regular GET requests for several resources.


lib/acme/client.rb
```diff
@@ -105,7 +105,7 @@ def new_order(identifiers:, not_before: nil, not_after: nil)
       identifiers
     else
       Array(identifiers).map do |identifier|
-        { type: 'dns', value: identifier }
+        identifier.is_a?(String) ? { type: 'dns', value: identifier } : identifier
```
unixcharles: Is this related to the post as get thing?

lib/acme/client.rb
```diff
@@ -151,13 +151,13 @@ def deactivate_authorization(url:)
...
   def challenge(url:)
-    response = get(url)
+    response = post_as_get(url)
     arguments = attributes_from_challenge_response(response)
     Acme::Client::Resources::Challenges.new(self, **arguments)
   end
 
   def request_challenge_validation(url:, key_authorization:)
-    response = post(url, payload: { keyAuthorization: key_authorization })
+    response = post(url, payload: {})
```
unixcharles: I don't understand why the key_authorization is being removed?

lib/acme/client/resources/challenges.rb
```diff
@@ -4,18 +4,14 @@ module Acme::Client::Resources::Challenges
...
   }
 
   def self.new(client, type:, **arguments)
-    klass = CHALLENGE_TYPES[type]
-    if klass
-      klass.new(client, **arguments)
-    else
-      { type: type }.merge(arguments)
-    end
+    (CHALLENGE_TYPES[type] || Unsupported).new(client, **arguments)
```
unixcharles: Doesn't seem related to post-as-get. What the use case for this?

lib/acme/client.rb
github1: It’s mentioned in this issue: 

https://github.com/letsencrypt/pebble/issues/192

Pebble will raise an error if key authorization is in the request payload:

https://github.com/letsencrypt/pebble/blob/master/wfe/wfe.go#L2135

lib/acme/client/resources/challenges.rb
github1: This is to gracefully handle receiving a TLS-ALPN challenge.  Basically this will fall back to a generic unsupported challenge type.

lib/acme/client.rb
github1: This is due to a separate change in pebble which I think ended up in 2.0.0 also.  The corresponding datatype in pebble seems to have changed from a string to a complex data type and I guess the deserialization logic changed.

https://github.com/letsencrypt/pebble/blob/master/wfe/wfe.go#L1528

https://github.com/letsencrypt/pebble/commit/bc4da68d49d78a38bfea869d38faadc8c714654c

lib/acme/client.rb
unixcharles: Thanks for the clarification

lib/acme/client/resources/challenges.rb
unixcharles: Fair enough

lib/acme/client.rb
unixcharles: OK, I see.

The idea for this API was to allow to pass identifier objects or string of dns.

For example,

```ruby
new_order(identifiers: 'foorbar.com')
new_order(identifiers: ['foorbar.com', 'www.foorbar.com'])
new_order(identifiers: {type: 'dns', value: 'www.foorbar.com'})
new_order(identifiers: [{type: 'dns', value: 'www.foorbar.com'}, {type: 'dns', value: 'foorbar.com'}])
new_order(identifiers: [{type: 'other-type', value: 'value'}])
```

Not that I look into it, I doubt the L104 ever did that https://github.com/unixcharles/acme-client/blob/1aea684e29ca61b7ec7d71f9517c786173af2b27/lib/acme/client.rb#L104 

It should probably have been something like this:

```ruby
identifiers = Array(identifiers)
payload['identifiers'] = if identifiers.all? { identifiers.is_a?(String) }
  identifiers.map do |value|
    { type: 'dns', value: value }
  end
else
  identifiers
end
```

github1: pushed a change to support this

github1:
@unixcharles do you want me to remove the `mode`/`kid` stuff?

unixcharles:
> @unixcharles do you want me to remove the mode/kid stuff?

This would change the method signature and force the user to update their code.

I will wait until we bump to a minor to avoid breaking change.

lib/acme/client.rb
```diff
@@ -101,13 +113,7 @@ def kid
...
   def new_order(identifiers:, not_before: nil, not_after: nil)
     payload = {}
-    payload['identifiers'] = if identifiers.is_a?(Hash)
-      identifiers
-    else
-      Array(identifiers).map do |identifier|
-        { type: 'dns', value: identifier }
-      end
-    end
+    payload['identifiers'] = self.class.prepare_order_identifiers(identifiers)
```
unixcharles: Why not just use a private instance method instead of pushing it up has a class method?

github1: I agree it seems kind of wrong being a class method. Something in the setup for `client` sends a `GET https://127.0.0.1/directory` request and I wanted to avoid setting up VCR. Is there an existing cassette I can use or a way to skip the http request?

unixcharles: IMO, it is just parameter handling. If you want to move it a private method and use `send(: prepare_order_identifiers )` like you would a public class method, I'm good with that.

unixcharles:
Publish in v2.0.5

PR #176, opened 2019-12-26
Extract connection initialization and wrap HTTP errors

weppos:
I was investigating some failures, and I found that the error `Connection reset by peer - SSL_connect` was somehow raised as Faraday exception:

```
Faraday::ConnectionFailed: Connection reset by peer - SSL_connect
```

and in other cases as Acme error:

```
Acme::Client::Error::Timeout
```

When I investigated I found that there are 3 places where a Faraday connection is initialized. One added a special middleware that also repackaged the exception

https://github.com/unixcharles/acme-client/blob/049b1838968b228dbf4fd5168fe02df25638224d/lib/acme/client.rb#L289-L293

but the others did not, such as https://github.com/unixcharles/acme-client/blob/master/lib/acme/client/resources/directory.rb#L70-L75

To offer a consistent experience, this PR introduces the following changes:

1. The Faraday middleware has been split into 2 middleware: one with Error management (that can be reused across all Faraday Acme instances), and one with Acme JWT features (that is only needed in certain requests)
2. The connection creation has been extracted and now new connections always use a single entry point for initialization

I am still trying to figure out a proper way to test these changes. The specs are passing and were kept unchanged during the refactoring. I'd like to add some extra specs, but I'm not sure if it's possible to simulate timeout or other connection errors using cassettes. I generally use webmock directly for that. I will continue to investigate.


spec/client_spec.rb
```diff
@@ -30,7 +30,7 @@
       }.to raise_error(Acme::Client::Error::BadNonce)
     end
 
-    it 'retry on bad nonce with bad_nonce_retry option', vcr: { cassette_name: 'nonce_retry' } do
+    it 'retry on bad nonce with bad_nonce_retry option', vcr: { cassette_name: 'nonce_retry' }, pending: "Is NONCE required for new_account?" do
```
weppos: @cpu I can probably figure this out myself by parsing the RFC, but I wonder if you have a quick answer to this question.

As a context, I am trying to determine if this is a good test case for the nonce or if I should pick another endpoint.

unixcharles:
> I am still trying to figure out a proper way to test these changes. The specs are passing and were kept unchanged during the refactoring. I'd like to add some extra specs, but I'm not sure if it's possible to simulate timeout or other connection errors using cassettes. I generally use webmock directly for that. I will continue to investigate.

I test everything with Pebble. For timeout, I think using webmock directly is the way to go.  You can't do that with VCR.

spec/client_spec.rb
cpu: @weppos `newAccount` does require a nonce. I believe all of the POST requests in RFC 8555 (including the POST-as-GET variant) require a nonce.

weppos: Thanks!

weppos:
@unixcharles it took years, but I finally managed to rebase this PR to present and apply the fix I had in mind when I started. I also managed to include tests to verify that we will continue to successfully capture Faraday connection issues.

Testing the changes was a challenge. I tried to use Faraday test adapter, and I explored many other options. Ultimately, using Webmock was the only reliable way.

I think this PR is now ready for review. If you feel like you still like the idea, and want to move it forward.

unixcharles:
its out in 2.0.14

weppos:
> Thanks @weppos, had forgot about this one 😃

Understandable, 2019 is a long time ago 😂 . Thanks for shipping it so quickly!

PR #179, opened 2020-02-26
Allow Faraday 1.0

### LINKED ISSUE #178: Allow Faraday 1.0.0
Many gems are starting to require Faraday 1.0.0, so they conflict with this gem.

Here is the Faraday 1.0.0 upgrade guide, it should not be too impactful for this gem I guess: https://github.com/lostisland/faraday/blob/master/UPGRADING.md

renchap:
Tests are passing locally, except one (that did not pass on `master`).

The change is trivial, but you my want to bump the major version, as this will make projects using this new version incompatible with other gems not allowing `faraday < 1.0.0`.

Fixes #178

### DIFF
--- a/acme-client.gemspec
+++ b/acme-client.gemspec
@@ -26,5 +26,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'vcr', '~> 2.9'
   spec.add_development_dependency 'webmock', '~> 3.8'
 
-  spec.add_runtime_dependency 'faraday', '~> 0.17', '< 1.0.0'
+  spec.add_runtime_dependency 'faraday', '>= 0.17', '< 2.0.0'
 end


renchap:
It also breaks Ruby < 2.3.
Maybe we could support change the dependency to `'>= 0.17', '< 2.0.0'` as this gem's code is compatible with both?
It will also avoid to worry about other gem's faraday requirements conflicting with this one.

zedtux:
@unixcharles can you please have a look at this PR?

acme-client.gemspec
```diff
@@ -26,5 +26,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'vcr', '~> 2.9'
   spec.add_development_dependency 'webmock', '~> 3.8'
 
-  spec.add_runtime_dependency 'faraday', '~> 0.17', '< 1.0.0'
```
unixcharles: I can't exactly recall why we added this strictness but if the tests pass I would rather just drop `'< 1.0.0'` and open up to `'>= 0.17' `.

renchap:
Here you go, updated the PR with `'>= 0.17', '< 2.0.0'`

Issue #174, opened 2019-12-11
Failing spec Acme::Client::CertificateRequest assigns the public key (FAILED - 1)


weppos:
I can't get the current test suite to work. I keep getting an error on the following spec:

```
Acme::Client::CertificateRequest
  assigns the public key (FAILED - 1)

Failures:

  1) Acme::Client::CertificateRequest assigns the public key
     Failure/Error: expect(request.csr.public_key.to_der).to eq(public_key_to_der(test_key))

       expected: "0Y0\x13\x06\a*\x86H\xCE=\x02\x01\x06\b*\x86H\xCE=\x03\x01\a\x03B\x00\x04-y\xF9y\x18\xE5E\xC0\x87\x99...\xA3+\xF6\x06\e\xC9{\x17%\x80\xCEI\x12+s\xC9\x06\xA6k\x8B\xB9P\xBC\x0FX,\x06\xF7\x0FwFp\xF9\x88\x9A"
            got: "0w\x02\x01\x01\x04 \x1C\xE3\xBF7\f\x96Q>\f$\xB4\xF8M!7\xED#d\xEF\xFD\xDE$\x900\xDAG=\xEC\xA1\xC1\xBE...\xA3+\xF6\x06\e\xC9{\x17%\x80\xCEI\x12+s\xC9\x06\xA6k\x8B\xB9P\xBC\x0FX,\x06\xF7\x0FwFp\xF9\x88\x9A"

       (compared using ==)

       Diff:
       @@ -1,2 +1,3 @@
       -0Y0*?H?*?H?=B-y?y?E???~?mU?7??6??L???????+?{%??I+s??k??P?X,?wFp???
       +0w ??7
              ?Q>
                 $??M!7?#d???$?0?G=????0?
       *?H?=?DB-y?y?E???~?mU?7??6??L???????+?{%??I+s??k??P?X,?wFp???

     # ./spec/certificate_request_spec.rb:101:in `block (2 levels) in <top (required)>'

Finished in 0.04114 seconds (files took 0.7615 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/certificate_request_spec.rb:98 # Acme::Client::CertificateRequest assigns the public key
```

Interestingly enough, the DER conversion seems to be causing the problem. I compared the PEM and it worked, but the DER does not.

```
(byebug) public_key_from_private(request.csr.public_key).to_pem == public_key_from_private(test_key).to_pem
true
[...]

PR #183, opened 2020-05-04
Update code to avoid Ruby 2.7.0 warnings

dgholz:
When running `acme-client` with Ruby 2.7.0, I see a few warnings like:
```
/Users/danielholz/.gem/ruby/2.7.0/gems/acme-client-2.0.6/lib/acme/client/resources/account.rb:8: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/danielholz/.gem/ruby/2.7.0/gems/acme-client-2.0.6/lib/acme/client/resources/account.rb:43: warning: The called method `assign_attributes' is defined here
/Users/danielholz/.gem/ruby/2.7.0/gems/acme-client-2.0.6/lib/acme/client/resources/order.rb:8: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/danielholz/.gem/ruby/2.7.0/gems/acme-client-2.0.6/lib/acme/client/resources/order.rb:49: warning: The called method `assign_attributes' is defined here
/Users/danielholz/.gem/ruby/2.7.0/gems/acme-client-2.0.6/lib/acme/client/resources/authorization.rb:8: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/danielholz/.gem/ruby/2.7.0/gems/acme-client-2.0.6/lib/acme/client/resources/authorization.rb:65: warning: The called method `assign_attributes' is defined here
/Users/danielholz/.gem/ruby/2.7.0/gems/acme-client-2.0.6/lib/acme/client/resources/challenges/base.rb:8: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/danielholz/.gem/ruby/2.7.0/gems/acme-client-2.0.6/lib/acme/client/resources/challenges/base.rb:43: warning: The called method `assign_attributes' is defined here
```

This PR updates the calls to `assign_attributes` to explicitly splat the argument.

I also removed the redefinition of `OpenSSL::SSL::VERIFY_PEER` in the spec helper, since it also printed a warning and didn't seem to affect tests.

There is one warning remaining related to the Faraday middleware; I've opened https://github.com/lostisland/faraday/pull/1153 to address that.

### DIFF
--- a/lib/acme/client/resources/account.rb
+++ b/lib/acme/client/resources/account.rb
@@ -5,7 +5,7 @@ class Acme::Client::Resources::Account
 
   def initialize(client, **arguments)
     @client = client
-    assign_attributes(arguments)
+    assign_attributes(**arguments)
   end
 
   def kid
--- a/lib/acme/client/resources/authorization.rb
+++ b/lib/acme/client/resources/authorization.rb
@@ -5,7 +5,7 @@ class Acme::Client::Resources::Authorization
 
   def initialize(client, **arguments)
     @client = client
-    assign_attributes(arguments)
+    assign_attributes(**arguments)
   end
 
   def deactivate
--- a/lib/acme/client/resources/challenges/base.rb
+++ b/lib/acme/client/resources/challenges/base.rb
@@ -5,7 +5,7 @@ class Acme::Client::Resources::Challenges::Base
 
   def initialize(client, **arguments)
     @client = client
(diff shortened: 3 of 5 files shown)


dgholz:
I also noticed that `Acme::Client::CertificateRequest assigns the public key` fails when I run the test suite locally (#174)

Issue #186, opened 2020-08-17
Support for fetching alternative certificates


gschorkopf:
### Problem

ACME client should support the ability to look up alternative certificates in order to continue using the IdenTrust chain during the yearlong gap (Sept 29 2020 to 2021) where Let's Encrypt will continue to cross-sign but are now using their own chain by default.

### Potential Solutions

1. Upgrading the [certificate](https://github.com/unixcharles/acme-client/blob/v2.0.5/lib/acme/client.rb#L130-L133) method to loop and download certificates until one matches the certificate authority expected. This is roughly what certbot is doing (see below)
2. Add a new method that will return all (including alternatives) certificates

### Notes

[Context on Transitioning to ISRG's root](https://letsencrypt.org/2019/04/15/transitioning-to-isrg-root.html)

There are additional details [here](https://community.letsencrypt.org/t/transition-to-isrgs-root-delayed-until-sep-29/125516/2) for Let's Encrypt's decision to:

> to start serving certificates with the “alternate” link relation to allow ACME clients to programmatically select a chain they prefer

There is a python implementation of this feature for [certbot here](https://github.com/certbot/certbot/pull/8080). Perhaps we could take a queue from the [tests they've written](https://github.com/certbot/certbot/pull/8080/files#diff-1b881d1b819f210cc99be0fc88e8fbe8R845).

In a world with many alternatives, it seems like you need to download them in order to understand how they operate. The links header only returns a URL for download with no additional information. The link header (in ruby) looks roughly like this:

```ruby
"link"=>
 {"index"=>"https://acme-v02.api.letsencrypt.org/directory",
  "alternate"=>"https://acme-v02.api.letsencrypt.org/acme/cert/longcert_id/1"}
```

I've done a bit of research. Today, the alternates (though there is only one) is the certificate that uses Let's Encrypt's chain. I imagine on Sept 29th the original and alternate certificates will swap (though have not confirmed this).

gschorkopf:
By the way, we are planning to do this work and contribute to this gem, hoping to use the work ahead of the Sept 29 deadline! Just wanted to make the issue publicly visible for y'all.

elct9620:
Oops, I miss this issue. I will subscribe to this issue for the update.

sawanoboly:
@unixcharles Hi, could you please your opinion on this Issue?

unixcharles:
You can use `force_chain` in `2.0.7`.

Thanks everyone

PR #187, opened 2020-08-19
fetches alternative_chains when option provided.

sawanoboly:
Hi, 

I've implemented for #186.

- If the preferred_chain argument is specified with the CA's CommonName, it will find for the CA and return a matching certificate chain.
- If there is no matching CA, it raises a NotFound exception

Usage:

```
> order.certificate  # returns default fullchain
=> "-----BEGIN CERTIFICATE-----\nMIIFQTCCBCmgAwIBAgITAPpzi...   # default certificate

# It using Array because it may be offered more than one in the future.
> order.certificate(preferred_chain: 'DST Root CA X3')  # returns a fullchain with 'DST Root CA X3' as the root
=> "-----BEGIN CERTIFICATE-----\nM...     
```

----

changes:

- remove fetch_alternative_chains option.
- use preferred_chain instead of `order#alternative_certificates`


lib/acme/client.rb
```diff
@@ -127,8 +128,16 @@ def finalize(url:, csr:)
...
   end
 
-  def certificate(url:)
+  def certificate(url:, fetch_alternative_chains: false)
     response = download(url, format: :pem)
+    if fetch_alternative_chains
+      @alternative_certificates = []
+      alt_chains_urls = fetch_alternative_links(response)
+      alt_chains_urls.each do |alt_chains_url|
+        response = download(alt_chains_url, format: :pem)
```
thommahoney, line 137: This re-assignment of `response` each time through the loop makes it impossible to get the original certificate response without invoking the method twice (once with `fetch_alternative_chains: false` and once with `fetch_alternative_chains: true`)

It also makes the value returned by this method unpredictable depending on the order of alternative certificates within the `Links` header.

As someone from the community using this library heavily, my preference would be to have a new method added (eg. `alternative_certificates`) and to have this existing method only modified to persist the `Links` in an instance variable, perhaps. This keeps the API compatible and separates the concerns a bit more.

lib/acme/client.rb
```diff
@@ -127,8 +128,16 @@ def finalize(url:, csr:)
...
+        @alternative_certificates.push response.body
+      end
+    end
     response.body
   end
```
ryansouza, line 142: I think it might be nicer to have a separate method entirely just for `get_all_certificates` or something. I think that storing a single response data on the client object itself makes it harder to explain/trace through whats going on. A new method returning an array wouldn't need to use state to avoid changing the existing methods like this

lib/acme/client.rb
sawanoboly: > persist the Links in an instance variable

Well, it's definitely better that way 👍  I'll try to change it to do so.

lib/acme/client.rb
sawanoboly: https://github.com/unixcharles/acme-client/pull/187#discussion_r478555219

As I said in the comment above, I'll try to set the `LInks` in an instance valiable, so I'll separate the methods accordingly. Thanks 😃

sawanoboly:
By the way, do you think you need to be able to determine the Issuer of an intermediate certificate like certbot?

e.g. In the case of certbot => `--preferred-chain "DST Root CA X3"`

IMO, I think that it is up to the user of the library to choose which intermediate certificates to use, so I believe it is sufficient for this library to get all the alternative_certificates.

sawanoboly:
@thommahoney @ryansouza 

Thank you for your comment 😃 I've updated the code.

lib/acme/client.rb
```diff
@@ -43,6 +43,7 @@ def initialize(jwk: nil, kid: nil, private_key: nil, directory: DEFAULT_DIRECTOR
     @kid, @connection_options = kid, connection_options
     @bad_nonce_retry = bad_nonce_retry
     @directory = Acme::Client::Resources::Directory.new(URI(directory), @connection_options)
+    @alternate_links ||= []
```
thommahoney, line 46: oh, whoops this is my mistake. I don't think you want to store these on the Client. Perhaps the `certificate` method on this class should return a tuple of `[response.body, alternate_links]` so that they can be stored on the Order class.

sawanoboly: I understand. I'll have to set the save destination to the Order class, or there will be a side effect.

sawanoboly:
@thommahoney 
Thanks for your comment 👍  I've updated return values of client.certificate and cleans up some changes before.

lib/acme/client/resources/order.rb
```diff
@@ -26,12 +27,20 @@ def finalize(csr:)
...
-      @client.certificate(url: certificate_url)
+      certificate, @alternate_links = @client.certificate(url: certificate_url)
+      certificate
     else
       raise Acme::Client::Error::CertificateNotReady, 'No certificate_url to collect the order'
     end
   end
 
+  def alternative_certificates
+    @alternate_links.map do |certificate_url|
```
thommahoney, line 38: one more bit of feedback here. If the `certificate` method hasn't been invoked yet, this method will return an empty array. I think the default value of the `@alternate_links` instance variable should be `nil` so that we can invoke the `certificate` method ourselves when it hasn't been called yet.

```suggestion
  def alternative_certificates
    _ = certificate if @alternate_links.nil?
    @alternate_links.map do |certificate_url|
```

thommahoney: Without this change, the caller must know to call the `certificate` method first, which isn't an ideal API in my opinon.

lib/acme/client.rb
unixcharles: > I think it might be nicer to have a separate method entirely just for get_all_certificates or something. I think that storing a single response data on the client object itself makes it harder to explain/trace through whats going on. A new method returning an array wouldn't need to use state to avoid changing the existing methods like this

I agree.

`#certificate(url: )` should work as expected assuming a URL pointing to an alternate. It's a bit of a weird return value.

You would end up with

```
_, alternatives = client.certificate(url: url)
certificate = client.certificate(url: alternative.first)
```

I would rather not change the existing API of the client. Especially method signature.

Could we go with something like

```
> client.alternative_certificates(url: certificate_url)
=> ["alternative_certificate_pem", ...]
> order. alternative_certificates
=> ["alternative_certificate_pem", ...]
```

Do you guys have a use case for fetching the certificate and the alternative from one method except for the fact that you already have it available? I'm assuming you would want one or the other.

lib/acme/client.rb
```diff
@@ -315,6 +316,18 @@ def fetch_chain(response, limit = 10)
...
   end
 
+  def fetch_alternative_links(response)
+    get_links(response, 'alternate').uniq.sort
+  end
+
+  # Attributes in the "Link" header such as `rel` are not unique (RFC8555).
+  # Therefore, it is possible that more than one alternative may be provided in the future.
+  # Currently the Link attribute is a single hash by Faraday::Middleware,
+  # but it will be returned as an Array to accommodate future changes.
```
unixcharles, line 326: Do you mean that `response.headers['link']` return a hash an sometime an array?

lib/acme/client/resources/order.rb
unixcharles: I agree. The design of the client is to hide the underlying http call from the user.

unixcharles (changes requested):

lib/acme/client.rb
ryansouza: I think the implementation is going to be:
- fetch the main cert and check root
- fetch alternatives looking for the wanted root

Like mentioned in another place I think the certbot API is nicer if we're okay using the x509 stdlib here to do some cert parsing to check the root cert common-name, instead of managing the main/alternative choice manually. That lets us ask the client to fetch us the IdenTrust certificate and it will automatically handle the switchover from main-cert to alternative-cert whenever LE swaps those two.

lib/acme/client.rb
sawanoboly: I've wrote about this as new issue.

https://github.com/unixcharles/acme-client/issues/188

alternate_certs should be considered for future use, as there may be more than one of them.

lib/acme/client.rb
sawanoboly: hmm...

for example, Should I implement the following?


```
$ order.certificate
#=> returns default certificate


$ order.certificate(preferred_chain: 'DST Root CA X3')  # option name come from certbot.
#=> returns 'DST Root CA X3' certificate

$ order.certificate(preferred_chain: 'not available chain name')
#=> raise error
```

sawanoboly: By the way, do you know a good way to compare the common name of the `OpenSSL::X509::Name` object?

```
> cert.issuer
=> #<OpenSSL::X509::Name CN=cackling cryptographer fake ROOT>


> cert.issuer.to_s
=> "/CN=cackling cryptographer fake ROOT"

> cert.issuer.to_a
=> [["CN", "cackling cryptographer fake ROOT", 12]]
```

It seems to me that #to_a is sufficient for now, but it's somewhat inaccurate.

sawanoboly:
```
> OpenSSL::X509::Certificate.new(xx)
=> #<OpenSSL::X509::Certificate
 subject=#<OpenSSL::X509::Name CN=Let's Encrypt Authority X3,O=Let's Encrypt,C=US>,
 issuer=#<OpenSSL::X509::Name CN=DST Root CA X3,O=Digital Signature Trust Co.>,
 serial=#<OpenSSL::BN 13298795840390663119752826058995181320>,
 not_before=2016-03-17 16:40:46 UTC,
 not_after=2021-03-17 16:40:46 UTC>

> OpenSSL::X509::Certificate.new(xx).issuer.to_a
=> [["O", "Digital Signature Trust Co.", 19], ["CN", "DST Root CA X3", 19]]

> OpenSSL::X509::Certificate.new(xx).issuer.to_a.flatten
=> ["O", "Digital Signature Trust Co.", 19, "CN", "DST Root CA X3", 19]
```

hmm...

sawanoboly: @ryansouza @unixcharles 
I've chaned an option to preferred_chain so that it will always return a single certificate chain. What do you think?

unixcharles: For me the comment still stand. `Acme::Client#certificate` returning is no a consistent `API`

`Acme::Client#order` return a order, `Acme::Client#order` return an account, `Acme::Client#authorization` return an authorization.

Now, `#certificate` would be returning an array of `[Certificate, [links...]]`, which would break the API and require a major release and make it less consistent. I would not expect a method called `#certificate` to have this type of return value.

sawanoboly: @unixcharles Thanks for comment 👍  I understand your opinion that we should avoid changing the API. 

So which of the following do you think is better?

- Add the preferred_chain option to `Acme::Client#certificate` to change it to return an alternate certificate directly
- Add the Acme::Client#altenate_certificate(preferred_chain: foobar) method and use it to get an alternate_certificate

unixcharles:
Opened https://github.com/unixcharles/acme-client/pull/190 based off this discussion.

Let me know if that implementation would work for you.

PR #191, opened 2020-09-12
Fix for HTTP Link header decoding

unixcharles:
* Move to Util and add spec
* Allow space between `;` and `rel=`, following https://tools.ietf.org/html/rfc8288#section-3-5
* Change hash value to an Array of values to allow multiple entries at the same rel

fix https://github.com/unixcharles/acme-client/pull/187/files#diff-e843384610a1a0cb8c21582bbed52188R323-R329

### DIFF
--- a/lib/acme/client/faraday_middleware.rb
+++ b/lib/acme/client/faraday_middleware.rb
@@ -82,18 +82,10 @@ class Acme::Client::FaradayMiddleware < Faraday::Middleware
     end
   end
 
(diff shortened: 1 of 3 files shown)


jhass:
I guess technically a `,` could appear between the `<>` or `""`? Maybe a `scan` based solution could be robust to that

unixcharles:
> I guess technically a `,` could appear between the `<>` or `""`? Maybe a `scan` based solution could be robust to that

Make sense. Should be more robust now.

PR #190, opened 2020-09-12
Add support for alternate certificate download

unixcharles:
This PR include

- A class to identify and match the chain
- Support for a `preferred_chain` optional karg on `Client#certificate`
- `preferred_chain` forward in `Order#certificate`

The only thing I'm not 100% is `preferred_chain` as a name. Its following certbot naming but it might confused people into thinking that it will try to find the prefer chain and fallback to something else if not found.

This implementation raise if the chain does not match. I think it would be up to the library user to implement fallback mechanism.

i.e.;

```
order.certificate(preferred_chain: 'IdenTrust') || order.certificate
```

PR #192, opened 2020-09-16
Add documentation on downloading an alternative certificate

chrisUsick:
Added some basic documentation with the  primary goal of demonstrating how to invoke `Order#certificate` with the `preferred_chain` keyword argument as per [this comment](https://github.com/unixcharles/acme-client/pull/190#issuecomment-692731843)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,17 +23,26 @@ gem 'acme-client'
 ```
 
 ## Usage
-* [Setting up a client](#setting-up-a-client)
-* [Account management](#account-management)
-* [Obtaining a certificate](#obtaining-a-certificate)
-  * [Ordering a certificate](#ordering-a-certificate)
-  * [Completing an HTTP challenge](#preparing-for-http-challenge)
-  * [Completing an DNS challenge](#preparing-for-dns-challenge)
-  * [Requesting a challenge verification](#requesting-a-challenge-verification)
-  * [Downloading a certificate](#downloading-a-certificate)
-* [Extra](#extra)
-  * [Certificate revokation](#certificate-revokation)
-  * [Certificate renewal](#certificate-renewal)
+- [Acme::Client](#acmeclient)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -189,6 +198,19 @@ end
...
+### Ordering an alternative certificate
+
+Let's Encrypt is transitioning to use a new, self-signed intermediate certificate. Starting September 29, 2020 new certificates will be signed by their own intermediate. To ease the transition on clients Let's Encrypt will continue signing an alternative version of the certificate using the old, cross-signed intermediate until September 29, 2021. In order to utilize an alternative certificate the `Order#certificate` method accepts a `preferred_chain` keyword argument. 
+For example, to download the cross-signed certificate after September 29, 2020, call `Order#certificate` as follows:
+ 
+```ruby
+order.certificate(preferred_chain: 'IdenTrust') || order.certificate
+``` 
+
+Note: if the specified peferred chain doesn't match an existing alternative certificate it will return `nil`.
```
gschorkopf, line 210: Typo

```suggestion
Note: if the specified preferred chain doesn't match an existing alternative certificate it will return `nil`.
```

gschorkopf: I think the current implementation is actually going to raise an error, and is depending on the consumer of the gem to catch `Acme::Client::Error::PreferredChainNotMatched`. Correct me if I'm wrong.

See https://github.com/unixcharles/acme-client/pull/190/files#diff-e843384610a1a0cb8c21582bbed52188R151

gschorkopf (changes requested):

chrisUsick:
Ah, you're right regarding the raising of the exception. Updated accordingly.

unixcharles:
I'm not too sure why but github auto closed this one when I merged other PR. I'll cherry pick the changes.

chrisUsick:
Okay, thanks!

PR #195, opened 2020-10-02
Update README to use the IdenTrust issuer name

keiko713:
Thank you for implementing the `force_chain` feature, really appreciate it!

I believe that `force_chain` takes the issuer name of the intermediate certificate, so I would like to update the example to use `DST Root CA X3`, which is the real issuer name of the IdenTrust cross signed intermediate certificate (https://letsencrypt.org/certificates/#cross-signing).

I've tested this with `DST Root CA X3` and it worked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,20 +200,20 @@ order.certificate # => PEM-formatted certificate
 
 ### Ordering an alternative certificate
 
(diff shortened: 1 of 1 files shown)


jhass (approved):

unixcharles (approved):

unixcharles:
Thanks

PR #197, opened 2020-11-10
Travis-ci: added support for ppc64le

sanjaymsh:
Hi,
I have added support for ppc64le build on travis-ci in the branch . The travis-ci build log can be tracked on the link :https://travis-ci.com/github/sanjaymsh/acme-client/builds/199804503 . I believe it is ready for the final review and merge.
Please have a look on it.

Thanks !!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+arch:
+  - amd64
+  - ppc64le
 language: ruby
 cache: bundler
 rvm:


jhass:
What's the benefit of this given this gem contains no platform dependent code itself?

sanjaymsh:
@jhass This package is included in the distros for ppc64le - we are trying to make sure that integration into the distros is pain free by making sure they have a clean bill of health from their development community. This simplifies the integration process for ppc64le based distros.

Thanks !!

jhass:
It would be nice if you took the time to answer the question rather than copy pasting your standard reply you posted to a hundred similar pull requests before :)

unixcharles:
I agree with @jhass here. This Gem is platform independent 

If it fail on ppc64le, there is nothing we could do about it, so I don't see the point of adding the build.

PR #199, opened 2021-02-10
Add padding to bring it up to the required byte size for EC keys

dschwar:
In the [RFC](https://tools.ietf.org/html/rfc7518#section-3.4) for JWA it says that R and S need to be padded to 32 bits (based on the key size).

I believe this is causing errors with lets encrypt of the form 
> Acme::Client::Error::Malformed : JWS verification error

copied from: https://github.com/nov/json-jwt/blob/d97f1be725dcb601fb8ca44b1d933614986c9c27/lib/json/jws.rb#L172


unixcharles (approved):

unixcharles:
Sorry about the delay. Thanks for the PR

PR #206, opened 2021-08-25
Fix compatibility with both Ruby 3.0 and Faraday 0.17.x

jhawthorn:
The gemspec claims compatibility with Faraday 0.17.x, but previously the test suite was not passing as the Faraday client built in `fetch_directory` did not have an adapter configured.

Additionally the test suite did not run under Ruby 3.0 (with any version of Faraday) because of a missing dependency on webrick.

Finally, when Faraday 0.17.x and Ruby 3.0 were used in combination `FaradayMiddleware` would error on initialization since Faraday 0.17 always passes an options hash, which are no longer compatible with kwargs.. I think it's safest to assume this as an options hash even in Faraday 1+ since that is relying on the `ruby2_keywords` hack, and that should keep compatibility with all versions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,10 @@ source 'https://rubygems.org'
 
 gemspec
 
+if faraday_version = ENV['FARADAY_VERSION']
+  gem 'faraday', faraday_version
+end
+
 group :development, :test do
   gem 'pry'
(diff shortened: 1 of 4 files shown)


unixcharles:
Thanks for the PR

unixcharles:
It out in 2.0.9

PR #220, opened 2023-01-31
Update test matrix

unixcharles:


.github/workflows/test.yml
```diff
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: ['2.6', '2.7', '3.0', truffleruby-head]
```
eregon, line 14: Was it intentional to remove truffleruby-head? (that's not EOL of coruse)
Was it failing? I couldn't find any failed build of it at https://github.com/unixcharles/acme-client/actions

unixcharles: I removed it because it was much slower than the other ruby and I was not really aware of anyone using it. The test suite was taking 1m30s vs 20s for c ruby.

If its useful to you and the truffle project, I don't mind adding it back.

eregon: I see. Yes testing of gems on truffleruby is very useful for us and we track to notice if/when it gets red.
I think one thing worth a try is to use a TruffleRuby release, as that's faster than truffleruby-head.
I did a run locally with 23.1.0 and `bundle exec rake spec` takes 12s and `bundle install` takes 10s.
I did a run in GitHub Actions at https://github.com/eregon/acme-client/actions/runs/6458950897/usage that's a bit slower, 1 minute total including bundle install from scratch. Specs themselves took 19s in https://github.com/eregon/acme-client/actions/runs/6458950897/job/17533733770. The `bundle install` is cached if it ran in the last 2 weeks.
So I made https://github.com/unixcharles/acme-client/pull/231.

PR #219, opened 2023-01-06
Add support for External Account Binding

saarons:
Happy to clean this up a bit but wanted to get eyes on this first pass.

PR #210, opened 2021-10-13
Add EAB support

scrazy77:
Add EAB support for ssl.com & zerossl.com.

PR #221, opened 2023-01-31
External account binding

unixcharles:
Based on #219 and #210

Add tests, tweak the interface and readme.

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -64,7 +64,7 @@ class Acme::Client
     if external_account_binding
       kid, hmac_key = external_account_binding.values_at(:kid, :hmac_key)
       if kid.nil? || hmac_key.nil?
-        raise ArgumentError, 'must specify kid or hmac_key key for external_account_binding'
+        raise ArgumentError, 'must specify kid and hmac_key key for external_account_binding'
       end
 
       hmac = Acme::Client::JWK::HMAC.new(Base64.urlsafe_decode64(hmac_key))
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -50,7 +50,7 @@ class Acme::Client
 
   attr_reader :jwk, :nonces
 
-  def new_account(contact:, terms_of_service_agreed: nil, eab_kid: nil, eab_hmac_key: nil)
+  def new_account(contact:, terms_of_service_agreed: nil, eab: nil, external_account_binding: nil)
     new_account_endpoint = endpoint_for(:new_account)
     payload = {
       contact: Array(contact)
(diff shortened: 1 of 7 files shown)


unixcharles:
@saarons @Shaeli do you mind giving it 👀

lib/acme/client.rb
```diff
@@ -123,7 +136,7 @@ def kid
     @kid ||= account.kid
   end
 
-  def new_order(identifiers:, not_before: nil, not_after: nil)
+  def new_order(identifiers:, not_before: nil, not_after: nil, eab: nil)
```
saarons, line 139: Why add `eab` here? I'm not sure I follow.

lib/acme/client.rb
```diff
@@ -50,7 +50,8 @@ def initialize(jwk: nil, kid: nil, private_key: nil, directory: DEFAULT_DIRECTOR
 
   attr_reader :jwk, :nonces
 
-  def new_account(contact:, terms_of_service_agreed: nil)
+  def new_account(contact:, terms_of_service_agreed: nil, eab: nil, external_account_binding: nil)
```
saarons, line 53: We could just force people to use `external_account_binding` instead of `eab`, I doubt it would change much.

saarons (commented):
Added 2 very minor comments but otherwise looks great to me.

lib/acme/client.rb
```diff
@@ -59,7 +60,19 @@ def new_account(contact:, terms_of_service_agreed: nil)
       payload[:termsOfServiceAgreed] = terms_of_service_agreed
     end
 
-    response = post(endpoint_for(:new_account), payload: payload, mode: :jws)
+    external_account_binding = eab || external_account_binding
+    if external_account_binding
+      kid, hmac_key = external_account_binding.values_at(:kid, :hmac_key)
+      if kid.nil? || hmac_key.nil?
+        raise ArgumentError, 'must specify kid or hmac_key key for external_account_binding'
```
Shaeli, line 67:
```suggestion
        raise ArgumentError, 'must specify kid and hmac_key key for external_account_binding'
```

Shaeli, line 67:
```suggestion
        raise ArgumentError, 'must specify kid and hmac_key key for external_account_binding'
```

Shaeli (commented):
Thanks a lot @unixcharles for adding the tests 🙇‍♀️ 

LGTM 👍 , added a very minor suggestion - feel free to ignore it

lib/acme/client.rb
unixcharles: No idea actually. I included it based on #210 but ended up deleting it after looking at [7.3.4 in spec](https://www.rfc-editor.org/rfc/rfc8555.html#section-7.3.4) I don't see any reference to it. I forgot to clean it up the argument.

lib/acme/client.rb
unixcharles: I agree

PR #223, opened 2023-02-22
Fix name error in documentation

Shaeli:
I was trying to rotate a private key following the documentation, it failed with:

```
/Users/Shaeli/git/github/octocerts/vendor/gems/ruby/3.1.0/gems/acme-client-2.0.13/lib/acme/client.rb:101:in 
`account_key_change': unknown keyword: :private_key (ArgumentError)
        from script/rotate_acme_client_key.rb:12:in `<main>'
```
because `private_key` is referenced instead of `new_private_key`

See [here](https://github.com/unixcharles/acme-client/blob/master/lib/acme/client.rb#L101) for the method definition

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ To change the key used for an account you can call `#account_key_change` with th
 ```ruby
 require 'openssl'
 new_private_key = OpenSSL::PKey::RSA.new(4096)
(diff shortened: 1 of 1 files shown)


unixcharles (approved):

unixcharles:
Thank you

PR #228, opened 2023-08-23
don't include dotfiles in gem

### LINKED ISSUE #227: acme-client's .rubocop.yml has a bad config value, causes error on CI
...this one is admittedly a little complex, and I'm not sure of the right fix, or if this repo is the place to fix it.

My project uses `acme-client`, and rubocop passes locally, but fails in CI, because the `acme-client` gem bundles its `.rubocop.yml`, which has this configuration:

```
Style/StringLiterals:
  Enabled: single_quotes
```

...which produces this error:

```
Error: Property Enabled of cop Style/StringLiterals is supposed to be a boolean and single_quotes is not.
RuboCop failed!
```

...which fails my build.

That cop configuration was introduced on [December 28 2015](https://github.com/unixcharles/acme-client/commit/17066bb93aa504172818c05a9bc39c709bb00da3#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5cR35-R36), so I'm guessing rubocop changed along the way, but I haven't tracked it down to a commit yet.

I see two simple solutions:
* Stop bundling `.rubocop.yml` into the `acme-client` gem (which also includes `.github`, `.gitignore`, and `.rspec`, though they don't seem to hurt anything)
* Update `acme-client`'s `.rubocop.yml` to configure `Style/StringLiterals` as `Enabled: true` and `EnforcedStyle: single_quotes`.

(Either of those should be a quick fix, though if another gem bundles its `.rubocop.yml` and has a similar misconfigured value, I'll still have a problem, so I'm looking into a way to configuring my rubocop on CI to avoid this.)

danbernier:
Fixes #227.

Here's the proof that it works:

```
dan@nexterday acme-client % git st
On branch exclude-rubocop-yml-from-gem
nothing to commit, working tree clean
dan@nexterday acme-client % gem build
WARNING:  open-ended dependency on bundler (>= 1.17.3, development) is not recommended
  if bundler is semantically versioned, use:
    add_development_dependency 'bundler', '~> 1.17', '>= 1.17.3'
WARNING:  open-ended dependency on webrick (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  See https://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: acme-client
  Version: 2.0.14
  File: acme-client-2.0.14.gem
dan@nexterday acme-client % mkdir gem_output && tar -xf acme-client-2.0.14.gem -C gem_output
dan@nexterday acme-client % cd gem_output
dan@nexterday gem_output % tar -xf data.tar.gz
dan@nexterday gem_output % ls -a .*
zsh: no matches found: .*
```

### DIFF
--- a/acme-client.gemspec
+++ b/acme-client.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = 'http://github.com/unixcharles/acme-client'
   spec.license       = 'MIT'
 
-  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) || f.start_with?('.') }
   spec.require_paths = ['lib']
 
   spec.required_ruby_version = '>= 2.3.0'


unixcharles:
I don't mind not packaging dotfiles with the gem in the future but imo this is something your project should fix. If you are vendoring your dependencies into your repository for whatever reason, have rubocop skip them.

danbernier:
Thanks!

>  If you are vendoring your dependencies into your repository for whatever reason, have rubocop skip them.

Agreed, but like I said [in a comment](https://github.com/unixcharles/acme-client/issues/227#issuecomment-1690486833) on the issue:

> I tried that, but it still finds the .rubocop.yml in there. My hypothesis is that the logic for finding config files is different from the logic for finding files to lint.

Yeah, though, it's kind of a hash of an issue. Thank you for accepting the PR despite that!

PR #233, opened 2024-01-16
Refactor http operation out of Directory

### LINKED ISSUE #232: acme-client fails when request google acme.
```
`nonces': undefined method `nonces' for nil:NilClass (NoMethodError)

      client.nonces
            ^^^^^^^
	from /Users/zack/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/acme-client-2.0.15/lib/acme/client/http_client.rb:143:in `store_nonce'
	from /Users/zack/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/acme-client-2.0.15/lib/acme/client/http_client.rb:82:in `on_complete'
```

Reason:

the response of `https://acme-staging-v02.api.letsencrypt.org/directory` comes without a Replay-Nonce header 
however `https://dv.acme-v02.test-api.pki.goog/directory`does which makes store_nonce fail. 

Example:

You don't need to obtain a real EAB.

```ruby

private_key = OpenSSL::PKey::RSA.new File.read 'private_key.pem'
client = Acme::Client.new(private_key: private_key, directory: 'https://dv.acme-v02.test-api.pki.goog/directory')

account = client.new_account(contact: 'mailto:zack@example.com',
                             terms_of_service_agreed: true,
                             external_account_binding:
                              { kid: 'xxxx',
                                hmac_key: 'xxxx' })

```

unixcharles:
- More consistent with other resources
- Fix quirks with using an http client that doesn't have an acme-client
- Fix crash when using Google ACME endpoint (because it return nonce on directory which the client doesn't know how to handle)
- 
fix #232

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -44,7 +44,7 @@ class Acme::Client
 
     @kid, @connection_options = kid, connection_options
     @bad_nonce_retry = bad_nonce_retry
-    @directory = Acme::Client::Resources::Directory.new(URI(directory), @connection_options)
+    @directory_url = URI(directory)
     @nonces ||= []
   end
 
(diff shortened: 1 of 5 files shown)


unixcharles:
Can you confirm this work for you @jimages?

jimages:
I have checked this fix work. Thanks for the awesome work!!!

PR #240, opened 2024-06-05
fix the encoding of ECDSA public keys with leading 00's

benburkert:
The public portion of an ECDSA public key has an 'x' and 'y' component. For a new account request that uses EAB token, this public key is added to the protected section of the JWS. When encoding the key, it's important to preserve all of the bytes in the different parts, even if they are zero bytes.

The previous encoding would dump the key to hex and split the x and y component from those, then encode them into an OpenSSL bignum, then dump that bignum to bytes and urlsafe-base64 encode that. But OpenSSL's bignum does not track the number of input bytes, only the most compact representation of the number. This means that leading zero bytes are dropped when re-encoded into a byte array from the bignum.

Pushing the 'x' and 'y' component through OpenSSL::BN is unecessary and causes issues for ECDSA keys that are randomly generated with leading zeros. This can be demonstrated with the following code:

``` irb
irb> OpenSSL::BN.new("00FF", 16).to_s(2)
=> "\xFF"
```

The test generates a key with a leading zero for the encoded 'x' component, then checks to make sure the JWS contains the corresponding 0x00 bytes in the encoding for 'x'.

### DIFF
--- a/lib/acme/client/jwk/ecdsa.rb
+++ b/lib/acme/client/jwk/ecdsa.rb
@@ -50,8 +50,8 @@ class Acme::Client::JWK::ECDSA < Acme::Client::JWK::Base
     {
       crv: @curve_params[:jwa_crv],
       kty: 'EC',
-      x: Acme::Client::Util.urlsafe_base64(coordinates[:x].to_s(2)),
-      y: Acme::Client::Util.urlsafe_base64(coordinates[:y].to_s(2))
+      x: Acme::Client::Util.urlsafe_base64(coordinates[:x]),
+      y: Acme::Client::Util.urlsafe_base64(coordinates[:y])
     }
   end
 
@@ -92,8 +92,8 @@ class Acme::Client::JWK::ECDSA < Acme::Client::JWK::Base
       hex_y = hex[2 + data_len / 2, data_len / 2]
 
       {
-        x: OpenSSL::BN.new([hex_x].pack('H*'), 2),
-        y: OpenSSL::BN.new([hex_y].pack('H*'), 2)
+        x: [hex_x].pack('H*'),
+        y: [hex_y].pack('H*')
       }
     end
   end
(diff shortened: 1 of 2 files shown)


benburkert:
This bug may affect encoded RSA keys too, I have not investigated that case.

unixcharles:
Thanks for the pull request @benburkert 

> This bug may affect encoded RSA keys too, I have not investigated that case.

I don't have enough context to known either but given that the change will apply to any type of keys. I'll need to do a bit of digging and bring myself up to date before merging. My brain completely purged all knowledge of `OpenSSL::BN` the minute I had it working.

Do you have a context where you found this to be necessary that I can use to test it out? I only use letsencrypt and never had any need to external account binding. I'm just confused as why it's only a problem now and would like to investigate.

Anyhow, I'll try to have it merged soon

benburkert:
> I don't have enough context to known either but given that the change will apply to any type of keys. I'll need to do a bit of digging and bring myself up to date before merging. My brain completely purged all knowledge of `OpenSSL::BN` the minute I had it working.

I can look into this in a bit.

> Do you have a context where you found this to be necessary that I can use to test it out? I only use letsencrypt and never had any need to external account binding. I'm just confused as why it's only a problem now and would like to investigate.

Right, you wouldn't run into this with LE since they don't use EAB tokens. ZeroSSL uses EAB tokens but this is not reproducible there (they likely zero pad if the number of bytes is short, [some libraries do this](https://ecdsa.readthedocs.io/en/latest/glossary.html#term-raw-encoding)). I work on anchor.dev and we've had this bug cause CI failures, and i'm pretty sure it's impacted users of https://github.com/anchordotdev/puma-acme. You can use anchor.dev or lcl.host to reproduce this, but the chances of randomly hitting this bug are 1/128. So it's not a new problem, just an unlikely one. But it can be reliably reproduced against anchor.dev if you setup the account with a "bad" private key [here](https://github.com/unixcharles/acme-client?tab=readme-ov-file#external-account-binding-support), I can provide instructions if you'd like.

PR #241, opened 2024-07-29
add base64 dependency for future-proofing

geemus:
When working in anchordotdev/puma-acme we started getting a failure against head ruby, because it no longer has base64 in stdlib, and instead you have to explicitly require it. You can see the error here:

https://github.com/anchordotdev/puma-acme/actions/runs/10150742999/job/28068516562

We added it ourselves to immediately address the issue, but wanted to pass the fix upstream to you as well. I don't believe it should change any current behavior, and should help future proof things so you won't need to address this when 3.4+ appear.

Just let me know if you have any questions or I can help further to land this. Thanks!


unixcharles:
Thanks for the heads up and the PR.

PR #247, opened 2025-03-18
Add validated to challenges

benburkert:
[RFC 8555, Section 8](https://datatracker.ietf.org/doc/html/rfc8555#section-8) includes "validated" as an optional field on Challenge objects. This PR adds a reader for the validated value to the `Base` challenge object.


unixcharles:
thank you

unixcharles:
Released in 2.0.21

PR #251, opened 2025-07-31
Add `orders_url` to Accounts

danbernier:
Acme::Client::Resources::Account doesn't expose the `orders_url` field from [section 7.1.2](https://www.rfc-editor.org/rfc/rfc8555#section-7.1.2). This PR adds it.

I added assertions for this to the existing tests (the values come from the existing cassettes). It seems unrelated to the tests that are there; let me know if you'd prefer the assertions split out.

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -286,6 +286,7 @@ class Acme::Client
       response.body,
       :status,
       [:term_of_service, 'termsOfServiceAgreed'],
+      :orders,
       :contact
     )
   end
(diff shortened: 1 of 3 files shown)


danbernier:
The failing specs are for order revocation, and JWK signing. These seem unrelated to my changes, and they pass locally (on ruby 3.1.4p223) - are these tests flakey, by any chance? I looked through other PRs but all their CI checks look like they were easy greens; their logs have expired, so I can't dig in further. I'm also unsure why some of the runs timed out.

Usually I'd rerun the failed jobs, but I don't have that option. @unixcharles, can you offer any insight into this?

unixcharles:
Released in `2.0.24`

danbernier:
Thank you, @unixcharles!

PR #253, opened 2025-08-07
Profiles Extension

benburkert:
Add support for [Automated Certificate Management Environment (ACME) Profiles Extension](https://datatracker.ietf.org/doc/draft-aaron-acme-profiles/). This ACME extension is supported by LetsEncrypt (both Boulder & Pebble). The directory resource can now include a `"profiles"` object that is accessible from the `Acme::Client#profiles` method, which returns a `Hash` of profile names to human-readable descriptions of the profile. A profile can be specified by name when creating an order with `Acme::Client#new_order`. The README has been updated with a section on usage for both.

Since the test cassettes were last recorded, LetsEncrypt's staging instance blocks emails from the `example.com` & `example.org` domains. To get around this, I've added the `EXAMPLE_DOMAIN` environment variable into some of the specs, and re-recorded some cassettes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,22 @@ new_private_key = OpenSSL::PKey::RSA.new(4096)
 client.account_key_change(new_private_key: new_private_key)
 ```
 
+### Profile Extension
+
+Provide a CA profile when creating a new order:
+
+```ruby
+order = client.new_order(identifiers: ['example.com'], profile: 'shortlived')
+```
+
+ACME servers may list supported profiles in the directory endpoint:
+
+```ruby
+client.profiles => {"classic": "https://example.com/docs/classic", "shortlived": "https://example.com/docs/shortlived"}
+```
+
+See the [RFC draft of certificate profiles](https://datatracker.ietf.org/doc/draft-aaron-acme-profiles/) for more info.
+
 ## Requirements
 
 Ruby >= 3.0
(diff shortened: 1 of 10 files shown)


unixcharles:
> Since the test cassettes were last recorded, LetsEncrypt's staging instance blocks emails from the example.com & example.org domains. To get around this, I've added the EXAMPLE_DOMAIN environment variable into some of the specs, and re-recorded some cassettes.

Fair enough. I usually record the test using Pebble (or Boulder when feature set trail behind) but honestly it's a real pain to setup. I think I might get more contributions with proper tests if I focus on recording with stating.

benburkert:
> > Since the test cassettes were last recorded, LetsEncrypt's staging instance blocks emails from the example.com & example.org domains. To get around this, I've added the EXAMPLE_DOMAIN environment variable into some of the specs, and re-recorded some cassettes.
> 
> Fair enough. I usually record the test using Pebble (or Boulder when feature set trail behind) but honestly it's a real pain to setup. I think I might get more contributions with proper tests if I focus on recording with stating.

Gotcha, pebble's default test config ([`./test/config/pebble-config.json`](https://github.com/letsencrypt/pebble?tab=readme-ov-file#binary)) includes the same profiles and should work for these tests, I can regenerate them if that's helpful.

PR #258, opened 2025-11-19
ARI related improvements

benburkert:
We implemented ARI in our internal fork of acme-client. I've been meaning to send a PR but in the mean time @nckslvrmn beat me to it (thanks!), so this is mostly just cosmetic changes that improve the API IMO. Before I get to those, the place I got stuck was how to test this using VCR testing. This is a extra challenge for ARI because the ARI workflow depends on orders & certs to be in the finished state before they can be successfully renewed, and there are also timing issues too. It looks like the recordings for the ARI tests added in #257 don't actually include the ARI information in the [payload](https://github.com/unixcharles/acme-client/pull/257/files#diff-5d2873a89a51d55ac6aadb0c38070eddbf47045bc97655d3c91b986f2aa5e50dR130)[1] or [response](https://github.com/unixcharles/acme-client/pull/257/files#diff-5d2873a89a51d55ac6aadb0c38070eddbf47045bc97655d3c91b986f2aa5e50dR162-R175). (Ditto for the new order request payload in [new_order_already_replaced.yml](https://github.com/unixcharles/acme-client/pull/257/files#diff-072d289a5d58a80e93b1be26fc69d65f85d75a9c5191057562ab43985359ecea).) So these tests could likely cause failures & confusion in the future, but I don't have a real solution, that is why I have marked some tests as "pending" in this PR.

This makes `RenewalInfo` reloadable by tracking the ARI ID value in the renewal info object. It also allows `Client.renewal_info` to accept an ARI ID as an alternative to the certificate object/data. And `Order` now tracks the ARI ID of the certificate the order is replacing in the `replaces` attribute. And finally, `Order#renew` is added as a helper method to easily replace a finished order.

[1] no `"replaces"` field present in the JWS' encoded `"payload"`:
```sh
$ echo 'eyJpZGVudGlmaWVycyI6W3sidHlwZSI6ImRucyIsInZhbHVlIjoiZXhhbXBsZS5jb20ifV19==' | base64 -D
{"identifiers":[{"type":"dns","value":"example.com"}]}
```

### DIFF
--- a/lib/acme/client.rb
+++ b/lib/acme/client.rb
@@ -224,13 +224,18 @@ class Acme::Client
     response.success?
   end
 
-  def renewal_info(certificate:)
-    cert_id = Acme::Client::Util.ari_certificate_identifier(certificate)
-    renewal_info_url = URI.join(endpoint_for(:renewal_info).to_s + '/', cert_id).to_s
+  def renewal_info(certificate: nil, ari_id: nil)
+    if certificate.nil? && ari_id.nil?
+      raise ArgumentError, 'must specify certificate or ari_id'
+    end
+
+    ari_id ||= Acme::Client::Util.ari_certificate_identifier(certificate)
+
+    renewal_info_url = URI.join(endpoint_for(:renewal_info).to_s + '/', ari_id).to_s
 
     response = get(renewal_info_url)
     attributes = attributes_from_renewal_info_response(response)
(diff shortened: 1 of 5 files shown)


unixcharles:
> gotcha, I think what I may do then is see if I can create a shell script that boots up pebble in a state where it's ready to run the ARI tests. That way the tests themselves won't be tied to pebble, and there will be a handy script for regenerating the VCR recordings.

You would be my hero.

PR #270, opened 2026-06-09
Avoid defining Object#valid_ip_address?

jeremyevans:
Looking at cfc29429faab0e278ee1260ad9375bab6549e984, it appears unintentional that this was defined as a top-level method (which makes it a private instance method of Object) and not a method of Acme::Client::CertificateRequest.


unixcharles:
Good catch. Didn't mean to pollute Object.

unixcharles:
Thank you for the PR. It's out in `2.0.32`.

